PR #56, opened 2014-05-08
Update GridFieldOrderableRows.php

pitchandtone:
If it's a many many relationship, then it should be based on View not edit, as the sort field will be on the relationship using many_many_extraFields. Plus the view permission allows the user to create a many relationship, so it should allow editing of the relationship.


pitchandtone:
Any more questions here? Can we merge?

nyeholt:
A couple of comments 

- can you bind getList to a variable prior to use? In some cases it can be a relatively costly operation, so would prefer it being called as few times as possible. 
- Can you use instanceof instead of is_a ?

pitchandtone:
Better?

PR #62, opened 2014-07-17
Allow dataless fields in inline add row component

mi3ll:
For things such as notes to the user etc.


nyeholt:
Only just saw this PR siting here... apologies! Anyway, the reason it wasn't committed in the first place was that if $fields has tabs or other composite fields, `->fieldByName` doesn't recurse into them to find the referenced field, causing the if block to be skipped. Not sure off the top of my head of a nice way to pick up dataless fields in this case?

mi3ll:
Ah, you have a point, I didn't notice this...

This pull request can be closed anyway I think, since I have found a walk around by using ReadOnlyField with noEscape enabled.

PR #79, opened 2015-01-18

pief:


pief:
Travis' build environment for SS master seems broken.

pief:
Ok GitHub is confusing me, opening separate pull requests for these issues.

PR #80, opened 2015-01-18

pief:


pief:
Travis build for SS master is currently broken.

pief:
Merged along with #81

PR #81, opened 2015-01-18
Extrasortfields

pief:
This allows for the specification of extra sort fields that are to be applied before the user-defined sort field's order. This is useful in scenarios where a GridField is used with one or more "fixed" columns which eg. categorize the entries and whose order is important but the user should still be allowed to modify the order within a category. To this account, the second change allows specifying a different column number for the reorder handles column.

More details in the commit messages.


pief:
Travis build for SS master is currently broken.

PR #84, opened 2015-02-13
FIX: Make sure inline adding to correct GridField

mi3ll:
I work with nested GridFields, and without this the inline button adds to all GridFields that are nested within one GridField.

### DIFF
--- a/javascript/GridFieldExtensions.js
+++ b/javascript/GridFieldExtensions.js
@@ -91,8 +91,9 @@
 				});
 			},
 			onaddnewinline: function(e) {
-				if(e.target != this[0])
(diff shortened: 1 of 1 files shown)


javascript/GridFieldExtensions.js
```diff
@@ -90,7 +90,10 @@
 					if(success) success.apply(grid, arguments);
 				});
 			},
-			onaddnewinline: function() {
+			onaddnewinline: function(e) {
+				if(e.target != this[0])
```
nyeholt: Can you please include the full parentheses for the if block ?

PR #85, opened 2015-02-23

pief:


pief:
This is basically the same as #80, of which I wrongly assumed that it got merged.

PR #103, opened 2015-09-10
Allowing to refer to external links from a modeladmin

spekulatius:
I ran into an issue where I wanted to simply add a link to a external source to a GridField. To achieve this issue I wrote the following code. It simply checks if there are is either a method or a field which provides an external link and adds it to the model admin under actions.

Please let me know if you are keen to add this.

Spek

### DIFF
--- a/code/GridFieldExternalLink.php
+++ b/code/GridFieldExternalLink.php
@@ -1,0 +1,67 @@
+<?php
+/**
+ * Displays a link to an external source referenced 'external link'
+ */
+class GridFieldExternalLink extends GridFieldDataColumns {
+
+	/**
+	 * Add a column for the actions
+	 *
+	 * @param type $gridField
+	 * @param array $columns
+	 */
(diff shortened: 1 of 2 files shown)


nyeholt:
What about another getter for ExternalLinkText ? And what about using $record->hasMethod() instead (so an extension could provide the implementation instead of the base record needing to)?

spekulatius:
Hello @nyeholt ,

Thank you for your review. I updated it to use hasMethod. That makes a lot of sense. Also I added the ability to adjust the text using a getter for ExternalLinkText.

Feel free to come back with further ideas.

Spek

spekulatius:
Thanks for merging this @nyeholt 

Do you know if there is a release planned anytime soon?

nyeholt:
I have a bunch of modules pending a tag and release, which will _hopefully_ be done later this week, possibly early next.

spekulatius:
Thanks for the update. I'll wait for this then.

PR #106, opened 2015-09-18
Add extension hook after re-ordering items

kinglozzer:
This is useful for things like cache invalidation, as the re-ordering uses `DB::query()` instead of the ORM so there’s no other way to hook into it.

Adding `parent::__construct()` is necessary - without that, extensions aren’t ever registered.


kinglozzer:
Any feedback on this? :)

nyeholt:
Heh, it wouldn't let me merge from phone in the first case, and it dropped off my inbox until then!

kinglozzer:
Thanks! :+1:

PR #108, opened 2015-09-28
Make travis test against 3.1 and 3.2 not master

dhensby:
This should get travis builds passing


dhensby:
Or not.. looks like the extensions won't work on 3.2, then?

PR #112, opened 2015-10-30
Add tests for 3 and 3.2 branches, support 5.6 and test master

tractorcow:
Since master doesn't support 5.3, this is added separately as a test vs php 5.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,20 @@
 language: php
 
-php:
+sudo: false
+
+php: 
   - 5.3
 
 env:
   - DB=MYSQL CORE_RELEASE=3.1
-  - DB=MYSQL CORE_RELEASE=master
 
+matrix:
+  include:
+    - php: 5.5
+      env: DB=MYSQL CORE_RELEASE=3
+    - php: 5.6
+      env: DB=MYSQL CORE_RELEASE=3.2
+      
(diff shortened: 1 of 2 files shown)


tractorcow:
Hm, master isn't any good, since composer restricts it. I've removed this for now.

Need to look into the 3.x / 3.2 errors.

tractorcow:
It looks like these broke because `ManyMany` is added as a new physical method in 3.2, so you can't really have relations named after that method anymore. ;)

tractorcow:
@nyeholt hi marcus, do you think you'd have a moment to look at this? This fix is necessary for 3.2 compatibility.

The code itself is fine, but the tests won't pass without these fixes. ;P

.travis.yml
```diff
@@ -1,12 +1,20 @@
 language: php
 
-php:
-  - 5.3
+sudo: false
+
+php: 
+ - 5.3
```
assertchris: 2 spaces to 1? SHAME ON YOU!

tractorcow:
Fixed @assertchris

nyeholt:
Thanks! Am a bit behind the 8 ball (and 9, and 10) at the moment :)

tractorcow:
Yay thanks!

PR #128, opened 2016-02-11
Fixed bug on editable column

SilbinaryWolf:
Fixed bug where HtmlEditorField would be placed inside editable column, causing the gridfield to have quirky behaviour that either broke the layout or wiped HtmlEditorField content from the DB on save.

### DIFF
--- a/code/GridFieldEditableColumns.php
+++ b/code/GridFieldEditableColumns.php
@@ -37,7 +37,7 @@ class GridFieldEditableColumns extends GridFieldDataColumns implements
 			$colRelation = explode('.', $col);
 			$value = $grid->getDataFieldValue($record, $colRelation[0]);
 			$field = $fields->fieldByName($colRelation[0]);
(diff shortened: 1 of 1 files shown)


nyeholt:
Will be looking at this in the next couple days, not ignoring it :) I've used an HTMLEditorField within an editable row before though without a problem, so is there a particular scenario where this is causing an issue?

SilbinaryWolf:
Hm. I used it with 'GridFieldOrderableRows' from the 'gridfieldextensions' module. Perhaps that broke something on my end.

SilbinaryWolf:
[http://i.imgur.com/lVOl2C3.png](url)

Still get the bug with GridFieldOrderableRows turned off. When I hit save, it bugs out the whole UI.

SilbinaryWolf:
Oh it's worth noting that I'm on Silverstripe 3.2.1 as well.

SilbinaryWolf:
Fixed in a different commit.

PR #136, opened 2016-03-12
FIX clickable editable rows

### LINKED ISSUE #135: Can't click grid rows
On most gridfields you can click anywhere on a grid row to edit a record. That doesn't work when using editablecolumns, but the highlight effect and pointer cursor still appear on hover creating a confusing user experience.

I would prefer if we could restore the clickable behaviour but if that can't be done for some reason I think we should remove the hover effect and pointer cursor.

jonom:
Intention here was apparently to stop the editing form loading when you click form controls added with GridFieldEditableColumns, but it was more aggressive than it needed to be. Now only prevents form load when you click form controls, clicking a regular cell will load the edit form as per standard gridfield behaviour (fixes #135)

### DIFF
--- a/code/GridFieldEditableColumns.php
+++ b/code/GridFieldEditableColumns.php
@@ -218,6 +218,9 @@ class GridFieldEditableColumns extends GridFieldDataColumns implements
 				));
 			}
 
+			// Add CSS class for interactive fields
+			if (!($field->isReadOnly() || $field instanceof LiteralField)) $field->addExtraClass('editable-column-field');
+
 			$fields->push($field);
 		}
 
--- a/javascript/GridFieldExtensions.js
+++ b/javascript/GridFieldExtensions.js
@@ -165,9 +165,12 @@
 		 * GridFieldEditableColumns
 		 */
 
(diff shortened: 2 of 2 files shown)


jonom:
@nyeholt any thoughts on this one? :)

nyeholt:
Sorry, have been plane hopping the last couple days :) Only thing I can think outside this is html editors and a dropdown with chosen applied - should they be explicitly added to the list of propagationStop things?

jonom:
No worries @nyeholt sorry for hassling you - "the squeaky wheel gets the grease" and all that :)

I hadn't considered those instances... best case scenario is any cell in a gridrow that has a field in it gets a 'field' css class added to it so we can target that generic selector instead of specifying a range of form elements. I'll have a poke around and see if that would be easy to do.

jonom:
I've pushed up a second commit with a more robust approach, if you like this I'll squash them.

nyeholt:
Yep looks good :) Squash and I'll merge.

jonom:
Squashed now, sorry for delay

PR #140, opened 2016-04-10
feat(GridFieldOrderableRows): Re-order gridfield items without immediate reload

SilbinaryWolf:
Added the ability to change setting on GridFieldOrderableRows so that instead of updating the order immediately, it updates when you hit 'Save Draft' or 'Save and Publish'.

The motivation behind doing this is to make reordering feel less painful on slower internet connections, slow/shared servers or when sorting versioned Data Objects (as the DB update is slower).

Not sure about the config/property name 'immediateUpdate' but that's the best I could think of for now.

Tested on Silverstripe 3.3.1

### DIFF
--- a/code/GridFieldOrderableRows.php
+++ b/code/GridFieldOrderableRows.php
@@ -10,7 +10,14 @@ class GridFieldOrderableRows extends RequestHandler implements
 	GridField_ColumnProvider,
 	GridField_DataManipulator,
 	GridField_HTMLProvider,
-	GridField_URLHandler {
+	GridField_URLHandler,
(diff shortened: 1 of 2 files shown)


nyeholt:
This looks like it may help with a lot of the 'unsaved relation list' issues as the write of the ordering happens after the initial record write occurs. Does that sound about right ?

SilbinaryWolf:
I don't think so. To be honest, I don't actually understand the 'Unsaved relation list' issues people are talking about, since try to use a gridfield on an unsaved DataObject, you'll get errors inside the subpages anyway?

PR #154, opened 2016-07-21
Bug with reordering sortable grid when some sort IDs are duplicated

dhensby:
My database has somehow entered a state where some sort orders are shared amongst different records.

This is causing the reordering to fail. This PR shows a proof in the form of a unit test.

I'm opening this as I believe the fix is fairly straight forward (just set the sort on all items in the list, instead of just the ones that have changes detected) but this may have unintended side-effects that I'm not aware of, so I wanted to open this and ask first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,36 @@
-language: php
+# See https://github.com/silverstripe-labs/silverstripe-travis-support for setup details
 
 sudo: false
 
-php: 
+language: php
+
+php:
   - 5.3
+  - 5.4
+  - 5.5
+  - 5.6
 
 env:
-  - DB=MYSQL CORE_RELEASE=3.1
+  - DB=MYSQL CORE_RELEASE=3.2
 
 matrix:
   include:
-    - php: 5.5
+    - php: 5.6
       env: DB=MYSQL CORE_RELEASE=3
     - php: 5.6
-      env: DB=MYSQL CORE_RELEASE=3.2
-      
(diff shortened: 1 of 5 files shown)


nyeholt:
So what is `$newOrder` actually coming out to in the test? I can see what you're saying, and the suggested fix sounds about right, though I wonder what the most performant approach to that will be.

dhensby:
@nyeholt this is the output from PHPUnit:

```
1) GridFieldOrderableRowsTest::testReorderItems
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => 5
-    1 => 4
-    2 => 3
-    3 => 2
-    4 => 1
+    1 => 3
+    2 => 4
+    3 => 1
+    4 => 2
 )
```

Which means the output is:

```php
array(
    0 => 5,
    1 => 3,
    2 => 4,
    3 => 1,
    4 => 2,
);
```

Any value with a matching sort value isn't being swapped. I think there's something going on with the way the logic is checking against position in an array instead of sort value?

The current code probably only gives the illusion of performance because if all items are in order and numbered in incremental order and an item is moved from bottom to top all values will need to change so we'll have an `UPDATE` per row.

What we could do instead is simply number them from 1 and increment them and just avoid an update if it's not changed, which will be expensive on a DB with sort numbers that are non-incremental, but from then on will be of similar performance to what we have now.

dhensby:
Fix added

dhensby:
thanks :)

PR #158, opened 2016-07-26
Bug where LastEdited is assumed to be on the same table as the sort

dhensby:
This demonstrates an error where the assumption is that the sort field is on the same table as `LastEdit` field.

In other words, the current assumption is that sort columns are always on the base table, which may not be the case.


SilbinaryWolf:
@dhensby Ping me when this is ready to be merged.

dhensby:
I'm trying to rebase this, but can't seem to get the tests to work and I don't really have the time to focus on this right now...

jonom:
@dhensby think was this fixed with https://github.com/silverstripe-australia/silverstripe-gridfieldextensions/pull/163 so can be closed?

dhensby:
@jonom it's fixed, but perhaps this proof should go in so that we don't get a regression in the future? I just need to spend some time trying to get the test rebased :/

jonom:
Oh my bad, missed the fact that it was a test rather than a fix

PR #160, opened 2016-08-18
BUG Ensure that unsaved items can be sorted

### LINKED ISSUE #159: Regression in 1.4 with userforms
Cross posted from https://github.com/silverstripe/silverstripe-userforms/issues/506

1.3 is fine. However, whenever saving a paged-form with 1.4 installed a new page is created each time, and the order gets messed up when drag-dropping.

Before:

image

Attemp to drag "Page 2" line one item down

image

It looks as though the drag succeeds, but then it sees the "First Page" with sort = 0, and that forces the item to the bottom.

tractorcow:
Fixes #159 (as a side effect)

Prior to this fix, any items created via GridFieldAddNewInlineButton would be unsortable. Now the template will include a sort grabber.

In order to ensure that the sort order is respected properly, the order is set via a hidden field, and is dynamically calculated on the fly with javascript, so that front end drag-dropping updates the record values.

A bit of work was done to ensure that the following components worked together:
- GridFieldEditableColumns
- GridFieldAddNewInlineButton
- GridFieldOrderableRows

Although there is a reference to GridFieldEditableColumns within GridFieldAddNewInlineButton, it's not actually necessary for it to be included in the same form for GridFieldAddNewInlineButton to work.

Likewise, GridFieldAddNewInlineButton will pick up the hidden field from GridFieldOrderableRows, and will use string replacement to rename the hidden field from '[GridFieldEditableColumns][0]' to '[GridFieldAddNewInlineButton][{%=o.num%}]' for use in the template.

Within GridFieldOrderableRows I've changed the behaviour so that the old 'order' posted property is no longer needed; Only the new hidden field will be necessary to post back. This does mean that in some cases, the SAME sort value may be saved twice by multiple components, but it is always consistent, and thus safe.

### DIFF
--- a/code/GridFieldAddNewInlineButton.php
+++ b/code/GridFieldAddNewInlineButton.php
@@ -101,7 +101,14 @@ class GridFieldAddNewInlineButton implements GridField_HTMLProvider, GridField_S
 
 				$content = $field->Field();
 			} else {
-				$content = null;
+				$content = $grid->getColumnContent($record, $column);
+
+				// Convert GridFieldEditableColumns to the template format
+				$content = str_replace(
+					'[GridFieldEditableColumns][0]',
+					'[GridFieldAddNewInlineButton][{%=o.num%}]',
+					$content
+				);
 			}
 
 			$attrs = '';
@@ -129,7 +136,10 @@ class GridFieldAddNewInlineButton implements GridField_HTMLProvider, GridField_S
 		}
 
 		$class    = $grid->getModelClass();
+		/** @var GridFieldEditableColumns $editable */
 		$editable = $grid->getConfig()->getComponentByType('GridFieldEditableColumns');
+		/** @var GridFieldOrderableRows $sortable */
+		$sortable = $grid->getConfig()->getComponentByType('GridFieldOrderableRows');
 		$form     = $editable->getForm($grid, $record);
 
 		if(!singleton($class)->canCreate()) {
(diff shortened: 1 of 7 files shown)


code/GridFieldAddNewInlineButton.php
```diff
@@ -101,7 +101,14 @@ private function getRowTemplate(GridField $grid, GridFieldEditableColumns $edita
 
 				$content = $field->Field();
 			} else {
-				$content = null;
```
tractorcow: This was the piece of code that was previously disabling the sort grip column.

code/GridFieldOrderableRows.php
```diff
@@ -450,10 +485,10 @@ protected function reorderItems($list, array $values, array $order) {
 			// *_versions table is updated. This ensures re-ordering works
 			// similar to the SiteTree where you change the position, and then
 			// you go into the record and publish it.
-			foreach(array_values($order) as $pos => $id) {
```
tractorcow: using array_values here to force contiguous sort values was making it impossible to insert new records in the same postback as existin records. Note that here $sortedIDs will have gaps for record sort values handled by GridFieldAddNewInlineButton (or others).

tractorcow:
Tested with:
 - GridFieldAddNewInlineButton, GridFieldEditableColumns, and GridFieldOrderableRows
 - GridFieldAddNewInlineButton and GridFieldEditableColumns only (no sorting)
 - GridFieldOrderableRows and GridFieldEditableColumns only (no inline add)
 - GridFieldOrderableRows only
 
Can't test GridFieldOrderableRows and GridFieldAddNewInlineButton only, because GridFieldAddNewInlineButton requires GridFieldEditableColumns to work.

code/GridFieldOrderableRows.php
tractorcow: In addition, array_values() forces the first $pos to be 0, which isn't good for checking if items have a sort set (=== 0 vs isset()).

tractorcow:
@nyeholt hey mate, eh, sorry if this is not a good time, but I'm in the middle of trying to get a release of userforms out, and I'd really like to get this into a fix so we can go forwards.

Is there any chance of getting this into a gridfieldextensions 1.4.2 tag please?

chillu:
I've peer reviewed this, with my limited understanding of how the sort logic works in detail. The general approach makes sense to me, I was able to reproduce the bug and the fix. The PR has no tests, but I guess unless we're going for Behat testing, there's not much point in testing the changed PHP logic since that's only supporting code for the JS changes that actually fix the bug. So a +0.75 from me ;)

PR #177, opened 2017-02-08
Initialising button before attempting to enable or disable it

JackOconnor21:
When trying to enable or disable buttons before running an initialisation for the button, an error will be thrown. I have just called the button function prior to enabling or disabling the button to combat this. I have also updated a class reference in GridFieldAddNewMultiClass.php as this was pointing to a non-existent class. I have added a use statement for Exception also.

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -20,6 +20,7 @@ use SilverStripe\ORM\ManyManyList;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\SS_Map;
 use SilverStripe\View\ViewableData;
+use Exception;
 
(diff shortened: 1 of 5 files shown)


src/GridFieldOrderableRows.php
```diff
@@ -20,6 +20,7 @@
 use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\SS_Map;
 use SilverStripe\View\ViewableData;
+use League\Flysystem\Exception;
```
SilbinaryWolf: This isn't used. Please remove.

oilee80:
Thanks @JackOconnor21 The JS was causing me issues 😉

PR #182, opened 2017-04-13
Enable namespace support for GridFieldAddNewMultiClass

### LINKED ISSUE #178: SS4 GridFieldAddNewMultiClass 'handleAdd' class getter
Currently the `handleAdd` function will take the param 'ClassName' from the passed $request. With SS4 this will just get the vendor from the namespace as opposed to Vendor\Project\Something\Class, meaning anything added with GridFieldAddNewMultiClass will get an error upon creation.

nikrolls:
This is for v1, but the same change will fix #178 if modified for `master`.

### DIFF
--- a/tests/GridFieldAddNewMultiClassWithNamespacesTest.php
+++ b/tests/GridFieldAddNewMultiClassWithNamespacesTest.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace SilverStripeAustralia\Test;
+
+use SapphireTest;
+use SS_HTTPRequest;
+use Form, Controller, FieldList;
+use GridField, GridFieldDetailForm, GridFieldAddNewMultiClass;
+
+class GridFieldAddNewMultiClassWithNamespacesTest extends SapphireTest {
+
+	public function testGetClassesWithNamespaces() {
+		$grid = new GridField('TestGridField');
+		$grid->setModelClass('SilverStripeAustralia\\Test\\NamespacedClass');
+
+		$component = new GridFieldAddNewMultiClass();
+
+		$this->assertEquals(
+			array(
+				'SilverStripeAustralia-Test-NamespacedClass' => 'NamespacedClass'
+			),
+			$component->getClasses($grid),
+			'Namespaced classes are sanitised'
+		);
+	}
+
+	public function testHandleAddWithNamespaces() {
+		$grid = new GridField('TestGridField');
+		$grid->getConfig()->addComponent(new GridFieldDetailForm());
+		$grid->setModelClass('SilverStripeAustralia\\Test\\NamespacedClass');
+		$grid->setForm(Form::create('test', Controller::create(), FieldList::create(), FieldList::create()));
+
+		$request = new SS_HTTPRequest('POST', 'test');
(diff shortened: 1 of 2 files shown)


dhensby:
looks like a good approach, but some minor code format issues.

tests/GridFieldAddNewMultiClassWithNamespacesTest.php
```diff
@@ -0,0 +1,60 @@
...
+		$this->assertInstanceOf('SilverStripeAustralia\\Test\\NamespacedClass', $record->getValue($response));
+	}
+
+}
+
+/**#@+
+ * @ignore
+ */
+
+class NamespacedClass {
```
robbieaverill: This should probably be in its own file identified as a "stub" and should implement the TestOnly interface so the manifest builder doesn't pick it up 👍

dhensby: Only `DataObject`s *need* to implement `TestOnly`

tractorcow: It doesn't hurt; Some classes that you don't want added to tests (E.g. those that extend phpunit classes) should be TestOnly to prevent class manifest trying to load classes with missing bases in production.

tractorcow:
Seems good enough to merge for v1. @nikrolls can you please follow up with another PR to fix these indentation errors? Maybe we'll fix on merge up to master otherwise.

pitchandtone:
@tractorcow since we already have a bunch of modules implementing this functionality, wouldn't it be best to move to convert as a static or similar:

reports/code/Report.php:
reports/code/ReportAdmin.php
silverstripe-admin/code/ModelAdmin.php

pitchandtone:
Can this go into master yet?

tractorcow:
All the files in master have moved, non-trivial merge. =( Someone have time to manually port up?

dhensby:
I'd do it but I'm not a member on these repos

tractorcow:
If someone cherry picks a PR I'll merge.

nglasl:
@dhensby, just having a look now, it says you're listed as an external collaborator for this module, however it's currently pending your response. Also just on these changes, can you please update the namespacing to the new `Symbiote`?

dhensby:
@nglasl I can't see it - can it be resent, please?

nglasl:
@dhensby, what happens if you go here? https://github.com/symbiote/silverstripe-gridfieldextensions/invitations

dhensby:
works :) thanks

dhensby:
merged up

colymba:
Seems classname needs to be sanitised on the handler too:
https://github.com/symbiote/silverstripe-gridfieldextensions/blob/master/src/GridFieldAddNewMultiClassHandler.php#L22?

tractorcow:
Yes, that's a bug.

tractorcow:
Fixed with https://github.com/symbiote/silverstripe-gridfieldextensions/commit/57acb83546e0bc411800092f9744d2849e756538

colymba:
thanks @tractorcow

PR #184, opened 2017-04-21
Added GridFieldSaveRow and GridFieldDuplicateRow

g4b0:
`GridFieldSaveRow` - allows to save inline edited record, also inside a ModelAdmin.
`GridFieldDuplicateRow` - allows to duplicate a record.

Fixes https://github.com/silverstripe-australia/silverstripe-gridfieldextensions/issues/89

### DIFF
--- a/code/GridFieldDuplicateRow.php
+++ b/code/GridFieldDuplicateRow.php
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * Duplicate the selected row
+ *
+ * @author Gabriele Brosulo <gabriele@brosulo.net>
+ * @creation-date 21-apr-2017
+ */
+class GridFieldDuplicateRow implements GridField_ColumnProvider, GridField_ActionProvider {
+    /* ******************************
+     * ** GridField_ColumnProvider **
+     * ******************************/
+
+    public function augmentColumns($gridField, &$columns) {
+        if (!in_array('Actions', $columns)) {
+            $columns[] = 'Actions';
+        }
+    }
+
+    public function getColumnAttributes($gridField, $record, $columnName) {
+        return array('class' => 'col-buttons');
+    }
+
+    public function getColumnMetadata($gridField, $columnName) {
+        if ($columnName == 'Actions') {
+            return array('title' => '');
+        }
+    }
+
+    public function getColumnsHandled($gridField) {
+        return array('Actions');
+    }
+
+    public function getColumnContent($gridField, $record, $columnName) {
+        if (!$record->canEdit())
+            return;
+
+        $field = GridField_FormAction::create(
+                        $gridField, 'DuplicateRow' . $record->ID, 'Clone', "duplicaterow", array('RecordID' => $record->ID)
+        );
+
+        return $field->Field();
+    }
+
+    /* *****************************
+     * **GridField_ActionProvider **
+     * ******************************/
+
+    public function getActions($gridField) {
+        return array('duplicaterow');
+    }
+
+    public function handleAction(GridField $gridField, $actionName, $arguments, $data) {
+
+        if ($actionName == 'duplicaterow') {
+
+            $model = $gridField->getModelClass();
+            $modified = $data[$model]['GridFieldEditableColumns'];
+
+            $obj = DataObject::get_by_id($model, $arguments['RecordID']);
+            $obj->duplicate();
+
+            // output a success message to the user
+            Controller::curr()->getResponse()->setStatusCode(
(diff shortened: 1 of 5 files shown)


g4b0:
Hey guys, any news about that? I'm working on a Javascript enchanting that handle LeftAndMain.CONFIRMUNSAVED message after saving a row, but if you don't intend to merge this PR I can avoid the effort of refactor it to fit into your module.

code/GridFieldSaveRow.php
```diff
@@ -0,0 +1,73 @@
...
+            $model = $gridField->getModelClass();
+
+            $obj = DataObject::get_by_id($model, $arguments['RecordID']);
+            foreach ($data[$model]['GridFieldEditableColumns'][$arguments['RecordID']] as $c => $v) {
+                $obj->{$c} = $v;
+            }
+            $obj->write();
+
+            // output a success message to the user
+            Controller::curr()->getResponse()->setStatusCode(
```
nyeholt: Is there a particular reason for explicitly setting a status code rather than letting the framework handle this?

code/GridFieldSaveRow.php
```diff
@@ -0,0 +1,73 @@
...
+        return array('saverow');
+    }
+
+    public function handleAction(GridField $gridField, $actionName, $arguments, $data) {
+
+        if ($actionName == 'saverow') {
+
+            $model = $gridField->getModelClass();
+
+            $obj = DataObject::get_by_id($model, $arguments['RecordID']);
```
nyeholt: Is there a particular reason for using get_by_id instead of something like `$item = $gridField->getList()->byID($arguments['RecordID']);` ?

code/GridFieldDuplicateRow.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+/**
+ * Duplicate the selected row
+ *
+ * @author Gabriele Brosulo <gabriele@brosulo.net>
+ * @creation-date 21-apr-2017
+ */
+class GridFieldDuplicateRow implements GridField_ColumnProvider, GridField_ActionProvider {
```
nyeholt: Does this essentially replicate https://github.com/unisolutions/silverstripe-copybutton/blob/master/code/GridFieldCopyButton.php ?

g4b0: Yes, it's the same. I didn't find it on addons.silverstripe.org because I was searching for "clone" or "duplicate".

code/GridFieldSaveRow.php
g4b0: I followed the official documentation for creating a gridfield actionprovider:
https://docs.silverstripe.org/en/3.1/developer_guides/forms/how_tos/create_a_gridfield_actionprovider/#basic-gridfieldcustomaction-boilerplate

Anyway, a simple `return true` should be sufficient?

code/GridFieldSaveRow.php
g4b0: No particular reason for that, I'm used to work with get_by_id, but if you think that `$gridField->getList()->byID($arguments['RecordID']);` is best for performance or stile we can use it.

g4b0:
Just before squashing my commits, I would like to ask if the workaroud I found to handle LeftAndMain.CONFIRMUNSAVED is good enough for you, since I don't like it a lot, but I can't find any way to do it otherwise. 

The problem is that after saving a row the entire gridfield is reloaded via AJAX, so entwine events are not attached to it. I solved by calling a custom function directly inside the onclick, it works but it's ugly :)

javascript/GridFieldExtensions.js
```diff
@@ -1,3 +1,8 @@
+function GFEremoveChanged() {
```
nyeholt: For entwine usage, it'll be something like 

```
$('css-selector-for-the-SaveRow-button').entwine({
    click: function (e) {

    }
});
```

Or more simply, you can just do it using a jquery "on" event handler

```
$(document).on('click', 'css-selector-for-the-SaveRow-button', function (e) {

});
```

micschk:
Hi, @nglasl why was this PR closed?
The issue is still unsolved (also in SS4.*) I think?

nglasl:
Thanks for raising this @micschk. I have a feeling I might know why this was closed, and it's likely because the `1` branch was deleted. Therefore, this PR no longer has a valid merge target. @g4b0, if this is still relevant, can you please open another PR? Alternatively, @micschk, you're welcome to take this on. It's going to require some tweaks to be SS4 compatible, but it was almost there based on conversations above!

PR #185, opened 2017-05-03
Fix SearchForm throwing a 404 on search

wilr:
The GridFieldAddExistingSearchHandler search has been converted to namespaces however this causes the FormAction to be incorrect and when performing a search, currently throws an error. The form name does not need to be namespaced.


nyeholt:
CI is busted at present because some upstream SS4 changes - will merge regardless for now.

PR #190, opened 2017-06-01
deny inline editing of records SS4

sachithra-gayan:
```
[Recoverable Error] Argument 1 passed to SilverStripe\Forms\Form::__construct() must be an instance of SilverStripe\Control\RequestHandler, instance of SilverStripe\GridFieldExtensions\GridFieldEditableColumns given, called in /var/www/html/straynew/gridfieldextensions/src/GridFieldEditableColumns.php on line 278 and defined
```


nyeholt:
So this fixes the issue you mentioned, or just disables it to prevent the error?

sachithra-gayan:
yes this issue can fix by this.

PR #194, opened 2017-06-25
fix to multi-class for ss4 support

pitchandtone:
Object create no longer exists


pitchandtone:
Although these are failing, it appears as though it's due to namespacing issues in core.

Issue #188, opened 2017-05-17
RFC: Improve sorting control for gridfield lists/pagination 


nicolewilliams16:
The problem: 
When using Gridfield, items can not be ordered with drag and drop if they exist across different pages.
 
While it is possible to set the items per page higher, this can instead lead to having too many items on one page, and needs developer input. It’s a binary approach which creates further issues eg a long list or having to go back for developer changes as the list grows 

Who has this problem: 
Raised by 2 public sector users

Agency 1- can not order a list of 11 intranet content categories alphabetically on live site as the list spreads over two pages  
Agency 2 - attempting to order lists of documents for download 

Proposed solution: 
Give greater control to sort the list view to CMS users, by allowing the view to be adjusted different numbers of items per page. 
 
This could be following approach of Google Analytics as per screenshots or by adding a free text field for more flexibility of items shown. 

![screen shot 2017-05-17 at 12 40 52 pm](https://cloud.githubusercontent.com/assets/11512277/26134513/54e7752c-3b03-11e7-81f2-cf8dd2cdc71f.png)
![screen shot 2017-05-17 at 12 40 47 pm](https://cloud.githubusercontent.com/assets/11512277/26134514/54e9a022-3b03-11e7-87d3-33bd9b66cd0b.png)
![screen shot 2017-05-17 at 12 40 40 pm](https://cloud.githubusercontent.com/assets/11512277/26134512/54e76dc0-3b03-11e7-953c-9e7f5b62f062.png)

wilr:
As long as developers have a way to specify the max number (i.e loading 5000 members on a single page might break servers) I'd love to see a component for it.

nicolewilliams16:
Thanks for the feedback @wilr, that suggests that the Google version with item number options is a better approach than free text then :)

michalkleiner:
Would it be more accurate saying that the problem is with 'moving items between pages in gridfields with more than one page'? That is the issue we were dealing with for our gov client.

Looking at it from that perspective, the solution could perhaps be to fix the mechanism for moving items across pages? Introducing a solution to dynamically adjust the number of items per page won't fix the broken functionality for moving items to the next or previous page.

As a workaround, we currently use https://github.com/ntd/silverstripe-togglepaginator to toggle the pagination, but that can obviously run into server resources issue with high numbers of items.

michalkleiner:
[...]

PR #201, opened 2017-07-31
NEW Add paginator with configurable page sizes

robbieaverill:
This PR adds a new "configurable paginator" component, which can be used to replace the default framework paginator. It provides a page size dropdown, as well as a cleaner public API to manage the state of the paginator.

When changing page sizes, the first shown record will be retained.

When page size changes are not compatible with the current paging controls (e.g. you're on page 7 of 7 and change to a large page size) the page numbers will be recalculated to reflect the new page size, but the first shown record will be retained on the screen.

This is a solution for the general consensus approach for #188

## Please note!

**This is a minor semver change, not a patch.** There's no `1` branch for the next minor `1.5` (?) release, so the branch will need to be changed to wherever is appropriate before merge!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ sudo: false
 
 language: php
 
+dist: precise
+
 php:
   - 5.3
   - 5.4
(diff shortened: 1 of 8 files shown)


nyeholt:
Looks good on the surface, can you add a note in the readme with a basic usage example?

robbieaverill:
@nyeholt sorry for the late reply. Yep, I'll push them up tomorrow.

robbieaverill:
Added, thanks :)

robbieaverill:
Thanks!

PR #206, opened 2017-08-25
BUG Fix incorrect drag-drop handle

tractorcow:
I've also copied across the most current indentation settings from asset-admin.

Screenshot:

image

### DIFF
--- a/templates/Symbiote/GridFieldExtensions/GridFieldOrderableRowsDragHandle.ss
+++ b/templates/Symbiote/GridFieldExtensions/GridFieldOrderableRowsDragHandle.ss
@@ -1,2 +1,2 @@
-<span class="handle"><span class="icon"></span></span>
+<div class="handle"><i class="icon font-icon-drag-handle"></i></div>
 $SortField
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,12 +15,6 @@ trim_trailing_whitespace = true
 [*.md]
 trim_trailing_whitespace = false
 
-[*.yml]
-indent_size = 2
-indent_style = space
-
-[{.travis.yml,package.json}]
-# The indent size used in the `package.json` file cannot be changed
-# https://github.com/npm/npm/pull/3180#issuecomment-16336516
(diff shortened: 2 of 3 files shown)


dhensby:
So we're using the sort icon as a drag-and-drop icon? Is there a more appropriate icon to use?

tractorcow:
I can't see one without modifying the font file, unless you find one you like.

dhensby:
Where are the font icons even coming from?

tractorcow:
`silverstripe` font that's provided in core.

Open up `/silverstripe-admin/client/src/font/icons-reference.html` in a local checkout to see the list preview. :)

sachajudd:
Currently working on a custom drag & drop icon @tractorcow @dhensby 🙂

tractorcow:
Hi @sachajudd do you think we could use something like https://github.com/primer/octicons instead of raster fonts? Our custom `silverstripe` font is based on a customised subset of these, but I think it wouldn't be too hard to extend this list in custom modules (e.g. silverstripe-extra-icons font).

tractorcow:
FYI we use http://fontastic.me/ to build the font files.

sachajudd:
Hey @tractorcow I'm going to add my icon to the font set that we use for the CMS 🙂

tractorcow:
Nice, let me know if you need any help. :)

tractorcow:
Also can use https://oblador.github.io/react-native-vector-icons/

image

templates/Symbiote/GridFieldExtensions/GridFieldOrderableRowsDragHandle.ss
```diff
@@ -1,2 +1,2 @@
-<span class="handle"><span class="icon"></span></span>
+<div class="handle"><i class="icon font-icon-sort"></i></div>
```
sachajudd: Icon class can be changed to font-icon-drag-handle once https://github.com/silverstripe/silverstripe-admin/pull/219 is merged 🙂

dhensby:
I've made those changes for you @sachajudd

PR #208, opened 2017-09-01
Merge up: 1.5 -> 2.0 -> 2 -> master

robbieaverill:
This PR is a merge up basically from having released a patch version in 1.5.x and 2.0.x, but since it also includes updating the configurable paginator for SS4 I didn't want to push it directly without review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ sudo: false
 
 language: php
 
+dist: precise
+
 php:
   - 5.6
   - 7.0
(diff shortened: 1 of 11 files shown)


robbieaverill:
The test failure is fixed by #209

nyeholt:
I think merging in Naomi's CSS updates has created a conflict - are you able to check ?

robbieaverill:
Thanks @nyeholt - I've merged upstream over the top

dhensby:
So I've gone over this and re-merged up. We're having problems with PHPUnit generating mocks which are compatible with splat operators `Injectable::create(...$arguments)`... I'm not sure how or if we can resolve this...

dhensby:
OK - I've checked locally and I don't have this problem, the method signature is correct :/

dhensby:
Issue was that it's using phpunit 3.7 8-|

update require-dev and we are good.

PR #213, opened 2017-09-15
BUG Fix inline editing / sortable

tractorcow:
Before:

image

After:

image

Also fixes crashing on error due to mis-matching of namespaced vs non-namespaced post keys.

Before:

image

After:

image


robbieaverill:
Better than broken 👍

Issue #170, opened 2016-11-11
GridFieldOrderableRows with Sqlite3 fails on reorder


cloeon:
Using SQLite3Database on SS 3.4.1, GridFieldOrderableRows fails on rows 471 and 506 respectively:

`$additionalSQL = (!$list instanceof ManyManyList) ? ', "LastEdited" = NOW()' : '';`

SQLite3 does not support _NOW()_.
Don't know if it is the best solution, but this works instead:

`$additionalSQL = (!$list instanceof ManyManyList) ? ', "LastEdited" = ' . DB::get_conn()->now() : '';`

robbieaverill:
I'm not sure whether this module advertises SQLite compatibility already, but I've triaged this as a feature enhancement and assumed it can be made compatible without breaking any APIs.

dhensby:
fixed by #216

PR #215, opened 2017-09-22
TEST Cover base table last edited update

dhensby:
Replaces #158 

A test to cover inherited classes where the sort column is not on the base table.

This was fixed by #163 but my test to prove it from #158 was never integrated.

### DIFF
--- a/tests/GridFieldOrderableRowsTest.php
+++ b/tests/GridFieldOrderableRowsTest.php
@@ -12,6 +12,8 @@ class GridFieldOrderableRowsTest extends SapphireTest {
 		'GridFieldOrderableRowsTest_Parent',
 		'GridFieldOrderableRowsTest_Ordered',
 		'GridFieldOrderableRowsTest_Subclass',
+		'GridFieldOrderableRowsTest_Unorderable',
+		'GridFieldOrderableRowsTest_OrderableChild',
 	);
 
 	public function testReorderItems() {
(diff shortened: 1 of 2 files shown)


dhensby:
also happens to fix #170

tests/GridFieldOrderableRowsTest.php
```diff
@@ -49,6 +51,29 @@ public function testReorderItems() {
...
+			$config
+		);
+		$originalOrder = $parent->Children()->column('ID');
+		$desiredOrder = array_reverse($originalOrder);
+		$this->assertNotEquals($originalOrder, $desiredOrder);
+		$_REQUEST['showqueries'] = true;
+		$reflection->invoke($orderable, $grid, $desiredOrder);
+		$newOrder = $parent->Children()->column('ID');
+		unset($_REQUEST['showqueries']);
+		$this->assertEquals($desiredOrder, $newOrder);
```
robbieaverill: The formatting of this test method was much nicer in #158 ;)

robbieaverill:
Does #163 fix #170, or does this PR? I see only test class changes in this

tests/GridFieldOrderableRowsTest.php
dhensby: white space added - and removed the bad `$_REQUEST` access! yikes

dhensby:
> Does #163 fix #170, or does this PR? I see only test class changes in this

This is just test coverage.

PR #216, opened 2017-09-22
Use Datetime::now() for last edited updates in OrderableRows

dhensby:
Use `SS_Datetime::now()` instead of `NOW()` in DB queries. This is better for testing.


dhensby:
fixes #170

PR #225, opened 2017-11-09
WIP: Add a new component for a pop-out 'meatball' menu

NightJar:
Which will allow per item (row) specific actions
such as grouping things like jumping to a specific tab
in the DetailForm, and if the displayed model is Versioned
then showing relevant (contextual) actions for Publishing,
Unpublishing, and Archiving (deleting) an item. This saves
a user from having to navigate through depths for each
record they wish to perform an action on, without crowding
the actions column for a row on the GridField.

Also included in this commit source structural updates
with regards to updating to a modern front end build
system consistent with the silverstripe/admin core.
This allows the module the advantage of being able to
include CMS centric components without having a hard
dependency to the project. This should cover a good
90%+ use cases, but allow for flexibility for framework
only builds wishing to use this module.


src/GridFieldMeatballMenuComponent.php
```diff
@@ -0,0 +1,311 @@
...
+
+    public function getColumnContent($gridField, $record, $columnName)
+    {
+        $this->buildDefaultActions($gridField, $record);
+
+        $templateData = ArrayData::create([
+            'Actions' => Convert::raw2json(array_values($this->getActions())),
+        ]);
+
+        return $templateData->renderWith(static::class);
```
robbieaverill: I know we made this change together, but this would mean child classes would have to define their own template, where they might just want to modify some behaviour. Perhaps we should stick to `__CLASS__` (not late bound) and let child classes overload this method if they want to, or projects could just modify the template in its own namespace in their theme

robbieaverill (changes requested):
One comment, plus would be good to add a blurb to `docs/en/index.md`

robbieaverill:
Things I've noticed (testing in combination with elemental):

* Choosing an element from the "add new" dropdown triggers a change on the field and prompts the save actions to highlight for the page
* Configurable paginator is fairly munted (styles) (logged separately as #227)

chillu:
Can we please not give APIs presentational names like "meatballs"? Yes, it's a funny name, but it doesn't communicate it's intent (I didn't know about it a month ago). And even if you know about it, that might not always be the UI component we choose to list those options. In a React/GraphQL powered UI, a GridField is just a set of loose UI constraints that can be translated by a frontend - e.g. "primary" and "secondary" action, rather than "popover after last button". Naming this component "more actions" seems more appropriate?

src/GridFieldMeatballMenuComponent.php
NightJar: Remember originally as we made the change it went to find a template given the class - reverting to this original state will allow for template hierarchy to be accounted into this selection, and child classes will fall back to this parent class' template.

Perhaps this would be better, allowing both flexibility and the static short hand.

robbieaverill:
>even if you know about it, that might not always be the UI component we choose to list those options.

That's true. More Actions seems like a good idea.

src/GridFieldMeatballMenuComponent.php
```diff
@@ -0,0 +1,311 @@
...
+use SilverStripe\Forms\GridField\GridFieldDetailForm_ItemRequest;
+use SilverStripe\Forms\Tab;
+use SilverStripe\Forms\TabSet;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\ArrayList;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\View\ArrayData;
+use SilverStripe\View\SSViewer;
+
+class GridFieldMeatballMenuComponent implements
```
chillu: Describe what it does.

tests/GridFieldMeatballMenuComponentTest.php
```diff
@@ -6,12 +6,12 @@
...
 use SilverStripe\Forms\Form;
 use SilverStripe\Forms\GridField\GridField;
-use Symbiote\GridFieldExtensions\GridFieldMeatballMenuComponent;
+use Symbiote\GridFieldExtensions\GridFieldActionsMenu;
 use Symbiote\GridFieldExtensions\Tests\Stub\ClassWithTabs;
 use Symbiote\GridFieldExtensions\Tests\Stub\VersionedClassWithTabs;
 use SilverStripe\Versioned\Versioned;
 
-class GridFieldMeatballMenuComponentTest extends SapphireTest
+class GridFieldActionsMenuComponentTest extends SapphireTest
```
robbieaverill: Filename is still meaty

nyeholt:
I'm new to the SS4 game, so from a documentation perspective, is there a defined standard approach to building JS for SS4 modules? Could that be linked to from the readme (or a contributing.md doc file)?

robbieaverill:
@nyeholt yeah there is (webpack with the silverstripe/webpack-config package and yarn for building). We've previously been giving lightweight examples and linking to the SS4 docs for it, which we can add to this module too.

docs/en/index.md
```diff
@@ -141,3 +141,44 @@ $paginator->setItemsPerPage(500);
...
+Actions Menu
+------------
+
+The `GridFieldActionsMenu` serves as an encapsulation type component, which for each item/row will render a pop-out/dropdown menu to the user - represented on in the UI as the 'meatball' type widget (as opposed to the ever popular 'hamburger' collapsing menu widget). The actions menu will render one of three classes of information:
+
+ - **Links to different edit screens**
+   If the item in the current row has an edit form (`getCMSFields`) which contains a root level `TabSet`, this division of items houses an entry for each of it's contained `Tab`s in the nature of e.g. `GridFieldEditButton`, etc. A parameter of `false` can be passed in to skip the first tab, in the case that clicking the row should invoke this action (such as the aforementioned edit button does).
+ - **Versioned actions**
+   Iff the item is versioned (via the `Versioned` extension), then applicable actions will be rendered in this division. These will be _Publish_, _Unpublish_, and _Archive_ (aka delete). It is worth noting that Publish and Unpublish are _not_ mutually exclusive, as a record can have changes ahead of it's published version (i.e. "Modified" as opposed to "Draft").
```
robbieaverill:
```diff
- Iff
+ If
```

robbieaverill (changes requested):

docs/en/index.md
NightJar: Nah, it's a thing :)
> In logic and related fields such as mathematics and philosophy, if and only if (shortened iff) is a biconditional logical connective between statements.

https://en.wikipedia.org/wiki/If_and_only_if

robbieaverill: Do you think it's appropriate to use in the context of this documentation and its audience?

NightJar: In developer documentation I'd have thought so, yes.

chillu (changes requested):
The "more options" dialog isn't keyboard accessible (tested in Chrome and Safari). I can focus on the icon itself and get it to expand, but then would expect up/down keys to work - or at least being able to tab into it. The dialog is added to the bottom of the form, which means after all other form fields. It has some ARIA declarations, so screen readers might actually pick this up correctly despite the weird ordering. Have you tested this? In general, you shouldn't use a tooltip for a dropdown UI control - Bootstrap separates these two use cases specifically, and you get all that accessibility work for free in those components. Compare this with the "preview mode" UI controls, which use a styled chosen.js dropdown. At some point they've regressed to not being keyboard focusable (I've made sure they were when we added the feature), but once you've expanded them keyboard navigation works just fine.

chillu:
This relates to https://github.com/silverstripe/silverstripe-admin/issues/287. @NightJar please make sure that you link any pull requests to their original storycards.

src/GridFieldActionsMenu.php
```diff
@@ -0,0 +1,315 @@
...
+     * Add each of the "Root" tabs to the actions for this component
+     *
+     * We expect that a tabbed list of fields will always have a singular root.
+     *
+     * @param DataObject $record
+     * @param callable $linkCallback
+     */
+    protected function addRootTabActions(DataObject $record, callable $linkCallback)
+    {
+        $tabSet = $record->getCMSFields()->first();
```
chillu: I don't think it's feasible to build the entire CMS fields for each record in order to auto-list tabs, in terms of performance impact here. Those could contain dozens of complex fields, conditional logic, record fetches. We need to make these views fast to make them useful, and this makes it harder going forward. 

How about switching this to explicit registration? So the GridField created for content blocks declares that every record has a group which happens deep link into tabs. The API for this component doesn't need to know anything about tabs. 

The component probably shouldn't even know about the difference between actions like "publish" and links like "settings". They're just actions that contain enough metadata (e.g. a deep link into a tab) to determine on the frontend how to handle them.

chillu:
@NightJar  There's no user feedback when triggering the publish option. At first I thought it was broken, until I noticed that the "publish" option was replaced with the "unpublished" option. Is this still planned? /cc @clarkepaul

NightJar:
@chillu yes it's planned. This is more of an MVP at the moment to my understanding. User feedback is a follow up story to my knowledge, along with adding little status tags per record (`draft`/`published`/`modified`), as the site tree does.

The usability issue stems from the use of a core component (seemingly designed for use as a menu), which on investigation also affects core: https://github.com/silverstripe/silverstripe-admin/issues/352

src/GridFieldActionsMenu.php
tractorcow: Do the "tab actions" even need to be added by default? Maybe just make this a capability, and allow users to manually addActiontoGroup() in their GridField_Config setup. You could eliminate all of this complexity.

Doing too much "just in case" the user may want it often becomes a chore to remove and replace those items anyway.

tractorcow (changes requested):

src/GridFieldActionsMenu.php
```diff
@@ -0,0 +1,315 @@
...
+     *
+     * @return array
+     */
+    public function getActions()
+    {
+        return $this->actions;
+    }
+
+    // The following 3 null return functions implement an undefined interface
+    // expected by GridFieldDetailForm_ItemRequest
```
tractorcow: This seems to be a symptom of mis-use of the API; You should pass in a GridFieldDetailForm to the GridFieldDetailForm_ItemRequest constructor, not an instance of this object.

Can you get the actual GridFieldDetailForm instance, and pass this in instead, in the code earlier in this file?

tractorcow (changes requested):

NightJar:
@nyeholt I've added notes about the front-end build chain to the README.md in the latest update :)

tractorcow:
One of the problems I've discussed with @NightJar are that this feature is still difficult to add custom links to, as well as remove any of the default actions from.

A quick win that I suggest to address this is to:
 - Move adding of "default actions" to the constructor. This allows users to construct this component, and then remove / modify any of the default actions at will.
 - Allow "Link" property for each action to take a callback as well as a string; This callback will take the $gridField and $record property, and return the final link to put into the template. This means that users can register these callbacks before the parent link is known, such as in DataObject::getCMSFields().
 
In the future I would like to see the ability to register custom request handler class names for specific actions, and allow HTTP POST actions for certain menu items. However, this can be a future enhancement.

clarkepaul:
@chillu @tractorcow @NightJar @robbieaverill Just to recap on conversations in case they get missed. Our implementation of the "more options" is supposed to use bootstrap dropdown component but instead we are using the popover component which won't have the right accessibility built-in for menus and actions. I requested it be built with the dropdown component but it wasn't available to be used and other areas use the popover instead, so that pattern continues to be followed unfortunately. 

If this is going to be rebuilt can we use the right bootstrap component? otherwise we are adding more debt.

tractorcow:
@NightJar why weren't we able to use dropdown?

NightJar:
@tractorcow It plain and simple does not work.

src/GridFieldActionsMenu.php
```diff
@@ -0,0 +1,277 @@
...
+        $actions = $this->getActions();
+        $actions = array_map($actionBuilder, $actions, array_keys($actions));
+        if (!$this->getShowFirstTab()) {
+            reset($actions);
+            $first = key($actions);
+            array_shift($actions[$first]);
+        }
+        // remove empty entries
+        foreach ($actions as $section => $sectionItems) {
+            $section = array_filter($sectionItems);
```
tractorcow: This variable is never used for anything; My IDE detects it as unused. Should something happen here?

src/GridFieldActionsMenu.php
```diff
@@ -0,0 +1,277 @@
...
+    /**
+     * If the object is versioned (has the {@link Versioned} extension applied) then add
+     * actions to publish, unpublish, and archive - depending on record state.
+     *
+     * @param DataObject $record
+     * @param string $typeName
+     *
+     * @return array
+     */
+    protected function addVersionedActions(GridField $gridField, DataObject $record, $typeName)
```
tractorcow: Doesn't match phpdoc

src/GridFieldActionsMenu.php
```diff
@@ -0,0 +1,277 @@
...
+    /**
+     * Add each of the "Root" tabs to the actions for this component
+     * We expect that a tabbed list of fields will always have a singular root.
+     *
+     * @param DataObject $record
+     * @param string $typeName
+     *
+     * @return array
+     */
+    protected function addRootTabActions(GridField $gridField, DataObject $record, $typeName)
```
tractorcow: Doesn't match phpdoc.

src/GridFieldRecordActionHandler.php
```diff
@@ -0,0 +1,138 @@
...
+        '$ID' => 'defaultAction',
+    ];
+
+    private static $default_action = 'detailForm';
+
+    protected $record = null;
+
+    public function __construct(GridField $gridField, DataObjectInterface $record)
+    {
+        $this->gridField = $gridField;
```
tractorcow: Property $gridField isn't defined

src/GridFieldRecordActionHandler.php
```diff
@@ -0,0 +1,138 @@
...
+
+    private static $url_handlers = [
+        '$ID//edit' => 'detailForm',
+        '$ID//view' => 'detailForm',
+        '$ID//ItemEditForm' => 'detailForm',
+        '$ID//$Action!' => '$Action',
+        '$ID' => 'defaultAction',
+    ];
+
+    private static $default_action = 'detailForm';
```
tractorcow: phpdoc for `@config` vars please. :)

src/GridFieldRecordActionHandler.php
```diff
@@ -0,0 +1,138 @@
...
+
+    protected $record = null;
```
tractorcow: PHPDoc.

tractorcow (changes requested):

NightJar:
Closing so there's no accidental merge. We're not ready for this yet, there are usability issues besides the PHP stuff.

src/GridFieldActionsMenu.php
NightJar: Hmm, there's supposed to be a deeper dereferencing there, indeed!

NightJar:
OK, so this is where we're at:

The PHP side functionality is functional, extensible, and peer reviewed. ✅ 
The react side CMS front-end interface is ... functional. ☑️ 
However the CMS interface work is not accessible. ❌ 

The last point is a deeper issue that requires a lot of work to either work around with a replication of the `silverstripe/admin` `PopoverField` to make it accessible, or to do this same work in `silverstripe/admin` to make a parallel field in order to not break the 'api' as such (which would require a major version bump), as other modules rely on the existing field. The other way to look at this would be to call it a bug and address it, but an argument can be made that we're using the wrong component as a base and semantics 'n all that...

Work on this meatball menu will continue at a later point since this is a non-blocker to operational use, and it ties in more closely to inline editing which will also begin at a later point.

chillu:
I'd suggest to re-open this PR, since it's impossible to continue peer review otherwise. We can always mark it as "changes requested". The GridField component still does too much, by calling `getCMSFields()` on every single record that gets listed in that GridField. I've outlined before why I think that's a bad idea. And have confirmed with Paul that it wasn't his UX intention to auto-add all tabs - he just wanted shortcuts in this one particular case of content blocks. There's things which you could argue are "bugs to be fixed later" (like the accessibility of the `PopoverField` component), but this isn't one of them: It sets expectation on a public API that we'd be committing to very soon as part of the CWP 2.0 recipe. So getting the PHP API right is a higher priority for me than accessibility concerns if this change should make it into CWP 2.0.

dhensby:
OK - the only way to reopen this is to reset the branch that was used for the PR to the latest commit in this PR and then re-open the PR then reset it back to the current commit. I'd do it, but I don't have push rights to CC repos (I thought I did 😞)

```sh
git checkout pulls/2.0/Menu-Encompassing-Action-Tasks
git reset --hard 274053a096d7fb3675833b6d59247d094b116493
git push -f
# [ go and open PR]
git reset --hard c5e9744280a82fc17db4b5ee79b6fc8b55d344fa
git push -f
```

chillu:
Same, no access on this repo. @robbieaverill @tractorcow does either of you?

robbieaverill:
@chillu you have owner permissions on the CCs org already, I've added @dhensby too

dhensby:
PR updated and re-opened, thanks @robbieaverill

src/GridFieldActionsMenu.php
NightJar: Fixed. Comment on line that hasn't changed, so doesn't "auto-resolve".

src/GridFieldActionsMenu.php
NightJar: Fixed. Comment on line that hasn't changed, so doesn't "auto-resolve".

tractorcow:
Good job thanks. :)

robbieaverill:
Marking title as "WIP" to hide it from our purr list.

clarkepaul:
Reactstrap is currently being added as well as an upgrade to Bootstrap 4 by @lukereative, maybe it works with that?

robbieaverill:
This has been replaced by MoreActions in core =)

PR #233, opened 2017-12-17
SS4 upgrade fixes.

mfendeksilverstripe:
* fixed incorrect reference to `Versioned` class
* fixed incorrect assumption that class names are the same as table names
* GridField reload action no longer triggers the change event as the changes are automatically saved by reorder action

As a result of these fixes this component is finally working with versioned items in the GridField.

### DIFF
--- a/javascript/GridFieldExtensions.js
+++ b/javascript/GridFieldExtensions.js
@@ -264,6 +264,82 @@
 		 */
 
 		$(".ss-gridfield-orderable tbody").entwine({
+      // reload the gridfield without triggering the change event
+      // this is because the change has already been saved by reorder action
+      reload: function (ajaxOpts, successCallback) {
+        var self = this.getGridField(), form = this.closest('form'),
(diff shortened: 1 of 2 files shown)


mfendeksilverstripe:
@dhensby Original code contains this:

```
$class = $list->dataClass();
if ($class == $this->getSortTable($list)) {
```

This doesn't work for any object that has class different from table name, which in SS4 is quite common. Of course, the issue is "only" with `Versioned` items as the `if` statement is used to detect `Versioned`.

The JS reformatting is a reasonable change requirement, I'll have a look at it on the next Hackday.

mfendeksilverstripe:
@dhensby I reverted the JS formatting changes. The functional JS changes should be much more clear now. Please review.

robbieaverill:
@mfendeksilverstripe sorry but would you mind rebasing onto 3.1, which is the latest suitable SS4 branch? Thanks :)

dhensby:
merges cleanly into 3.1 - no rebase needed :)

PR #250, opened 2018-03-06
Can't reorder many many versioned classes

deracs:
Currently when updating a list of a many many versioned items the handleReorder method was going through the ORM process. Previously it had a `$isVersioned = false` variable set that would allow for this to go through the non ORM way. 

I've reverted back to how it previously was to fix this issue


robbieaverill:
I think I'm OK with this change, seeing as #243 originally had this logic, and the merge up of #233 is what I think caused this regression.

PR #251, opened 2018-03-14
sanitise classname on addmulticlass handler

andrewandante:
Backport of fix that is in the 3 branch


dhensby:
`composer require symbiote/silverstripe-gridfieldextensions:1.5.x-dev` :P

PR #253, opened 2018-03-27
Use DBQuery to update the stage & live sort field

deracs:
There was an issue for versioned objects not saving to the live table. So updated it to use DBQuery.
Also needed to detect if the the list was a many_many for the first executeOrder condition.

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -2,6 +2,7 @@
 
 namespace Symbiote\GridFieldExtensions;
 
+use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\ClassInfo;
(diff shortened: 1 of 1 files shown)


src/GridFieldOrderableRows.php
```diff
@@ -531,24 +537,21 @@ protected function reorderItems($list, array $values, array $sortedIDs)
...
-            $baseTable = DataObject::getSchema()->baseDataTable($list->dataClass());
-
+            $baseTable = $sortTable;
+            if (class_exists($sortTable)) {
+                $baseTable = singleton($sortTable)->baseTable();
+            }
             $isBaseTable = ($baseTable == $sortTable);
             if (!$list instanceof ManyManyList && $isBaseTable) {
-                $additionalSQL = ", \"LastEdited\" = '$now'";
+                $additionalSQL = ', "LastEdited" = NOW()';
```
dhensby: Why have you moved from using `$now` to `NOW()`, this is a regression in behaviour...

robbieaverill:
Thanks for contributing @deracs! Feel free to reopen this PR if you want to update it with @dhensby's feedback

PR #256, opened 2018-04-26
Inherited sort column on versioned data object reorder corrected

mfendeksilverstripe:
Corrected a case when sort column is located in a table that belongs to ancestor class of the item we are trying to reorder.

Fixes following setup:

`BaseItem` (is `Versioned`) - has the `Sort` column
`Item` (extends `BaseItem` and is `Versioned`) - is used as items in the list within `GridField`

Reorder action with the following setup would not detect that item is versioned.

### DIFF
--- a/tests/GridFieldOrderableRowsTest.php
+++ b/tests/GridFieldOrderableRowsTest.php
@@ -9,8 +9,10 @@ use SilverStripe\Forms\GridField\GridFieldConfig_RelationEditor;
 use Symbiote\GridFieldExtensions\GridFieldOrderableRows;
 use Symbiote\GridFieldExtensions\Tests\Stub\StubOrderableChild;
 use Symbiote\GridFieldExtensions\Tests\Stub\StubOrdered;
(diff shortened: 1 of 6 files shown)


mfendeksilverstripe:
@robbieaverill Added a unit test to cover the case that this changeset is trying to fix. Please review.

tests/GridFieldOrderableRowsTest.php
```diff
@@ -125,5 +132,80 @@ public function testGetSortTable()
...
+        foreach ($parent->MyHasManySubclassOrderedVersioned() as $item) {
+            $item->publishRecursive();
+        }
+
+        // there should be no difference between stages at this point
+        $differenceFound = false;
+        foreach ($parent->MyHasManySubclassOrderedVersioned() as $item) {
+            /** @var  StubSubclassOrderedVersioned|Versioned $item */
+            if ($item->stagesDiffer()) {
+                $differenceFound = true;
```
dhensby: you could `$this->fail()` here

dhensby:
merged by hand and changed the fail condition I commented on

Issue #25, opened 2013-09-02
Missing license


simonwelsh:
There's no license for this module. Please include one, so people can use it :)

PR #258, opened 2018-05-11
FIX Work with ManyManyThroughList

### LINKED ISSUE #254: GridFieldOrderableRows - No support for ManyManyThroughList
There is no support in `GridFieldOrderableRows` for `ManyManyThroughList`. Note that `ManyManyThroughList` doesn't extend `ManyManyList`. 

We need to find all cases where a `ManyManyList` is handled and handle `ManyManyThroughList` in that place as well.

dhensby:
This PR depends on https://github.com/silverstripe/silverstripe-framework/pull/8065

This PR adds support for sorting via a `ManyManyThroughList`

fixes #254

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -19,6 +19,7 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectInterface;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\ManyManyList;
+use SilverStripe\ORM\ManyManyThroughList;
 use SilverStripe\ORM\Map;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\FieldType\DBDatetime;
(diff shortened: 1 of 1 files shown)


dhensby:
There are other instances in the class where we check if `$list instanceof ManyManyList` but none of those seemed to need changing to make sorting in the CMS work.

mfendeksilverstripe:
@dhensby I tested your new changes on a setup which uses versioned `Many Many Through Relation` and it doesn't work properly. I noticed two problems here:

* GridFieldOrderableRows::executeReorder() has some custom logic for `ManyManyList` and is likely in need of custom logic for `ManyManyThroughList` as well (note that `ManyManyThroughList` does not inherit from `ManyManyList`), the issue I found is that the `$current` contains empty values
* GridFieldOrderableRows::reorderItems() handles versioned `ManyManyThroughList` as non-versioned which is definitely not good, have a look at how the `$isVersioned` is determined, you may need to add a new case for `ManyManyThroughList`

NightJar:
@dhensby @mfendeksilverstripe - I have an in progress work of this as a solution to https://github.com/symbiote/silverstripe-gridfieldextensions/issues/25 - however there needs to be a bit more done in order to have it work _well_, thus there's been no PR just yet. If one or both of you could check it out I'd be grateful.

tractorcow:
Merged https://github.com/silverstripe/silverstripe-framework/pull/8065, so this PR can be re-reviewed.

NightJar:
Still too shallow an approach, as @mfendeksilverstripe mentioned.

dhensby:
replaced by #260

PR #260, opened 2018-05-20
NEW Add support for ManyManyThrough relations

### LINKED ISSUE #252: Orderable Rows not Sorting many_many relationship with extraFields
I have a many_many relationship with extraField `SortOrder`. 

My gridfield is a `GridFieldConfig_RelationEditor` with `addComponent(new GridFieldOrderableRows('SortOrder')` set. 

However when I tried to drag to sort my relations, the records reordered notification did pop
image
But the database `SortOrder` field do not change accordingly. 

Both the DataObjects are versioned, with the new SS4 versioned extension. 

EDIT: Only fix was to use https://github.com/UndefinedOffset/SortableGridField.

### LINKED ISSUE #254: GridFieldOrderableRows - No support for ManyManyThroughList

NightJar:
Previously relationships defiend as `many_many` came in a special type of `RelationList` (`ManyManyList`) - however now this can be one of two types of `RelationList` depending on the type of definition, with both being valid `many_many` relationships (i.e. `ManyManyThrough`).

This had the unfortunate side effect of seeing the OrderableRows component in (the least) cease functioning correctly.

This also addresses the inability to version many_many releationships (i.e. `many_many_extraFields` - e.g. ordering), as the join table itself was not versioned. Using a `many_many through` allows for this to be versioned via the intermediary join object, as with any other `DataObject`.

Resolves #254 and resolves #252

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -148,6 +148,8 @@ class GridFieldOrderableRows extends RequestHandler implements
      * Checks to see if the relationship list is for a type of many_many
      *
      * @param SS_List $list
+     *
+     * @return bool
      */
     protected function isManyMany(SS_List $list)
     {
@@ -493,7 +495,7 @@ class GridFieldOrderableRows extends RequestHandler implements
      */
     protected function executeReorder(GridField $grid, $sortedIDs)
     {
-        if (!is_array($sortedIDs)) {
+        if (!is_array($sortedIDs) || empty($sortedIDs)) {
             return false;
         }
         $sortField = $this->getSortField();
@@ -540,6 +542,7 @@ class GridFieldOrderableRows extends RequestHandler implements
             $toRelationName = $manipulator->getLocalKey();
             $sortlist = DataList::create($joinClass)->filter([
                 $toRelationName => $items->column('ID'),
+                // first() is safe as there are earlier checks to ensure our list to sort is valid
                 $fromRelationName => $items->first()->getJoin()->$fromRelationName,
             ]);
             $current = $sortlist->map($toRelationName, $sortField)->toArray();
@@ -583,19 +586,19 @@ class GridFieldOrderableRows extends RequestHandler implements
         //   - Related item is versioned...
         //       - Through object is versioned: handle as versioned
         //       - Through object is NOT versioned: THROW an error.
-        $isManyMany = $this->isManyMany($list);
-        if ($isManyMany && $list instanceof ManyManyThroughList) {
+        if ($list instanceof ManyManyThroughList) {
             $listClassVersioned = $class::create()->hasExtension(Versioned::class);
             // We'll be updating the join class, not the relation class.
             $class = $this->getManyManyInspector($list)->getJoinClass();
             $isVersioned = $class::create()->hasExtension(Versioned::class);
 
+            // If one side of the relationship is versioned and the other is not, throw an error.
             if ($listClassVersioned xor $isVersioned) {
                 throw new Exception(
                     'ManyManyThrough cannot mismatch Versioning between join class and related class'
                 );
             }
-        } elseif (!$isManyMany) {
+        } elseif (!$this->isManyMany($list)) {
             $isVersioned = $class::create()->hasExtension(Versioned::class);
         }
 
@@ -712,12 +715,12 @@ class GridFieldOrderableRows extends RequestHandler implements
         }
 
         if ($this->isManyMany($list)) {
-            $intropector = $this->getManyManyInspector($list);
+            $introspector = $this->getManyManyInspector($list);
             $extra = $list instanceof ManyManyList ?
-                $intropector->getExtraFields() :
-                DataObjectSchema::create()->fieldSpecs($intropector->getJoinClass(), DataObjectSchema::DB_ONLY);
-            $key   = $intropector->getLocalKey();
-            $foreignKey = $intropector->getForeignKey();
+                $introspector->getExtraFields() :
+                DataObjectSchema::create()->fieldSpecs($introspector->getJoinClass(), DataObjectSchema::DB_ONLY);
+            $key   = $introspector->getLocalKey();
+            $foreignKey = $introspector->getForeignKey();
             $foreignID  = (int) $list->getForeignID();
 
             if ($extra && array_key_exists($this->getSortField(), $extra)) {
@@ -739,7 +742,7 @@ class GridFieldOrderableRows extends RequestHandler implements
      * these functions are moved to ManyManyThroughQueryManipulator, but otherwise retain
      * the same signature.
      *
-     * @param ManyManyList|ManyManyThroughList
+     * @param ManyManyList|ManyManyThroughList $list
      *
      * @return ManyManyList|ManyManyThroughQueryManipulator
      */
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -2,6 +2,7 @@
 
 namespace Symbiote\GridFieldExtensions;
 
+use Exception;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\ClassInfo;
@@ -17,9 +18,11 @@ use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectInterface;
+use SilverStripe\ORM\DataObjectSchema;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\ManyManyList;
-use SilverStripe\ORM\Map;
+use SilverStripe\ORM\ManyManyThroughList;
+use SilverStripe\ORM\ManyManyThroughQueryManipulator;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\FieldType\DBDatetime;
 use SilverStripe\Versioned\Versioned;
@@ -142,6 +145,16 @@ class GridFieldOrderableRows extends RequestHandler implements
     }
 
     /**
+     * Checks to see if the relationship list is for a type of many_many
+     *
+     * @param SS_List $list
+     */
+    protected function isManyMany(SS_List $list)
+    {
+        return $list instanceof ManyManyList || $list instanceof ManyManyThroughList;
+    }
+
+    /**
      * Sets extra sort fields to apply before the sort field.
      *
      * @param string|array $fields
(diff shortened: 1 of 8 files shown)


NightJar:
OK, it's a bit mess, but it seems to work :)
The mess can be addressed at another time: #261

thats4shaw:
I gave this a test as per #252 @NightJar.

Seems to work pretty well for me.  The only hiccup I had was not versioning the joiner model initially.  I'm also assuming in this scenario that the `SortOrder` field should just be defined in the joiner class rather than use `$many_many_extraFields`.

NightJar:
Hi @thats4shaw, thanks for giving this a go! Yes you're right, it should now be defined on the intermediary model - I've not looked into whether or not that would work with `extraFields`, but I assume not.

I inserted some errors to be thrown on the case of mis-matched versioning strategies between the intermediary and the related model - but if I'm honest it's mostly because it hurt my brain trying to think through what it would mean if I didn't. I figure if it remains or becomes an issue it could be dealt with separately (in another issue).

src/GridFieldOrderableRows.php
```diff
@@ -141,6 +144,16 @@ public function getExtraSortFields()
         return $this->extraSortFields;
     }
 
+    /**
+     * Checks to see if the relationship list is for a type of many_many
+     *
+     * @param SS_List $list
+     */
```
robbieaverill: Missing return type

src/GridFieldOrderableRows.php
```diff
@@ -507,47 +530,73 @@ protected function executeReorder(GridField $grid, $sortedIDs)
...
                 }
             }
+        } elseif ($items instanceof ManyManyThroughList) {
+            $manipulator = $this->getManyManyInspector($list);
+            $joinClass = $manipulator->getJoinClass();
+            $fromRelationName = $manipulator->getForeignKey();
+            $toRelationName = $manipulator->getLocalKey();
+            $sortlist = DataList::create($joinClass)->filter([
+                $toRelationName => $items->column('ID'),
+                $fromRelationName => $items->first()->getJoin()->$fromRelationName,
```
robbieaverill: What happens if `->first()` returns null?

src/GridFieldOrderableRows.php
```diff
@@ -507,47 +530,73 @@ protected function executeReorder(GridField $grid, $sortedIDs)
...
+        //       - Through object is versioned: handle as versioned
+        //       - Through object is NOT versioned: THROW an error.
+        $isManyMany = $this->isManyMany($list);
+        if ($isManyMany && $list instanceof ManyManyThroughList) {
+            $listClassVersioned = $class::create()->hasExtension(Versioned::class);
+            // We'll be updating the join class, not the relation class.
+            $class = $this->getManyManyInspector($list)->getJoinClass();
+            $isVersioned = $class::create()->hasExtension(Versioned::class);
+
+            if ($listClassVersioned xor $isVersioned) {
```
robbieaverill: Is this implying that the base class and a MMT that is owns can't both be versioned?

NightJar: No, the opposite.

robbieaverill: "if the list class is versioned or (never and) the data class versioned"

I'm confused

NightJar: If one is versioned and the other is not, throw an error.
I'll add a comment :)

src/GridFieldOrderableRows.php
```diff
@@ -507,47 +532,75 @@ protected function executeReorder(GridField $grid, $sortedIDs)
...
-                break;
+        // Model doesn't have sort column because sort column is on ManyManyThroughList...
+        //   - Related item is not versioned:
+        //       - Through object is versioned: THROW an error.
+        //       - Through object is NOT versioned: handle as NOT versioned
+        //   - Related item is versioned...
+        //       - Through object is versioned: handle as versioned
+        //       - Through object is NOT versioned: THROW an error.
+        $isManyMany = $this->isManyMany($list);
+        if ($isManyMany && $list instanceof ManyManyThroughList) {
```
ScopeyNZ: `if($isManyMany)` is completely redundant here.

src/GridFieldOrderableRows.php
```diff
@@ -648,10 +715,13 @@ protected function getSortTableClauseForIds(DataList $list, $ids)
             $value = '= ' . (int) $ids;
         }
 
-        if ($list instanceof ManyManyList) {
-            $extra = $list->getExtraFields();
-            $key   = $list->getLocalKey();
-            $foreignKey = $list->getForeignKey();
+        if ($this->isManyMany($list)) {
+            $intropector = $this->getManyManyInspector($list);
```
ScopeyNZ: `$introspector` or `$inspector`.

NightJar: lol 😅

src/GridFieldOrderableRows.php
```diff
@@ -648,10 +715,13 @@ protected function getSortTableClauseForIds(DataList $list, $ids)
...
+            $extra = $list instanceof ManyManyList ?
+                $intropector->getExtraFields() :
+                DataObjectSchema::create()->fieldSpecs($intropector->getJoinClass(), DataObjectSchema::DB_ONLY);
```
robbieaverill: Multi line ternary isn't the best for readability, perhaps you could make this an if statement? Failing that, putting the `?` and `:` at the start of lines helps to indicate that it's part of a ternary condition and not just bad indentation

src/GridFieldOrderableRows.php
```diff
@@ -667,4 +737,27 @@ protected function getSortTableClauseForIds(DataList $list, $ids)
 
         return "\"ID\" $value";
     }
+
+    /**
+     * A ManyManyList defines functions such as getLocalKey, however on ManyManyThroughList
+     * these functions are moved to ManyManyThroughQueryManipulator, but otherwise retain
+     * the same signature.
+     *
+     * @param ManyManyList|ManyManyThroughList
```
robbieaverill: Missing variable name

tests/OrderableRowsThroughTest.yml
```diff
@@ -0,0 +1,40 @@
+Symbiote\GridFieldExtensions\Tests\Stub\ThroughDefiner:
+  DefinerOne:
+    ID: 1
+  DefinerTwo:
+    ID: 2
```
robbieaverill: Defining IDs in fixtures isn't a very good idea - you'd ideally let the DB driver assign them via auto increments. Maybe just define a simple varchar for the sake of fixturing instead?

tests/OrderableRowsThroughVersionedTest.php
```diff
@@ -0,0 +1,107 @@
...
+    protected static $extra_dataobjects = [
+        ThroughDefiner::class,
+        ThroughIntermediary::class,
+        ThroughBelongs::class,
+    ];
+
+    protected static $required_extensions = [
+        ThroughDefiner::class => [ Versioned::class ],
+        ThroughIntermediary::class => [ Versioned::class ],
+        ThroughBelongs::class => [ Versioned::class ],
```
robbieaverill: Can you remove the spacing around the array values?

tests/OrderableRowsThroughVersionedTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $config = new GridFieldConfig_RelationEditor();
+        $config->addComponent($orderable);
+
+        // This test data is versioned - ensure we're all published
+        $parent->publishRecursive();
+        // there should be no difference between stages at this point
+        foreach ($parent->Belongings() as $item) {
+            if ($item->stagesDiffer()) {
+                $this->fail('Unexpected diference found on stages');
+            }
```
robbieaverill: Perhaps you could switch this if statement for `$this->assertFalse($item->stagesDiffer(), 'No difference found on stages');` (note typo)

tests/OrderableRowsThroughVersionedTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $this->assertNotEquals($originalOrder, $desiredOrder);
+
+        $reflection->invoke($orderable, $grid, $desiredOrder);
+
+        $newOrder = $parent->Belongings()->sort($sortName)->map($sortName, 'ID')->toArray();
+
+        $this->assertEquals($desiredOrder, $newOrder);
+
+        // pass forward to testReorderedPublish
+        return $parent;
```
robbieaverill: I don't think is happening, and if it were I think I'd suggest it'd be better not to couple unit tests together. I see you've declared `testReorderedPublish` as depending on this test, but it doesn't actually use its return value - can you remove

tests/OrderableRowsThroughVersionedTest.php
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+        $this->assertTrue($differenceFound);
+
+        $parent->publishRecursive();
+
+        // there should be no difference between stages anymore
+        foreach ($parent->Belongings() as $item) {
+            if ($item->stagesDiffer()) {
+                $this->fail('Unexpected diference found on stages');
```
robbieaverill: Same comment re: `assertFalse($item->stagesDiffer(), ...`

tests/OrderableRowsThroughVersionedTest.php
NightJar: Sorry, can't do. It is passed in at line 86, and used on line 90, 98, & 101 (in `testReorderedPublish`). I agree that normally it's not a good idea to string tests along like this, but in this case I feel it's acceptable in that if the first test fails there's no point in testing the second as it truly is a prerequisite (in order to publish one must be able to save) - I'm not aware of any method to fixture version information.

It could also be argued that this could all be part of one big test if this dependency is the case, but I feel it is cleaner this way.

NightJar: Added docblock `@param` for `testReorderedPublish` to compliment the comment above the return there and make it a bit clearer :)

robbieaverill: Ah I see, the return value is passed into testReorderedPublish as an arg. Ok

gorriecoe:
When is this likely to be tagged?  This fixes a major issue for our team so ideally this week if possible?

robbieaverill:
@gorriecoe here you go: https://github.com/symbiote/silverstripe-gridfieldextensions/releases/tag/3.2.0

PR #272, opened 2018-09-28
FIX Orderable rows now respects actual MMTL sort orders instead of incrementing from SiteTree

### LINKED ISSUE #268: Many Many Through sort column value after reorder with Page object
Module version: `3.2.0`

### Testing setup:

`ParentObject` has a many many through relationship to `Page` via `RelationObject`.

`Sort` column is defined on `RelationObject` and also on `Page` (inherited from `SiteTree`).

### Issue

`ParentObject` has a gridfield which is used to manage the MMT relationship. Add a bunch of pages to the relation and re-order them. This will work but the `Sort` value will not be as expected.

Normally, the `Sort` value starts from `1` and as you add items it increments (`2`, `3`...). The next value should be computed as a `max` from all the items in the relation. This is how it's working currently in most cases, but not this one. The difference is that the `Sort` column is defined in both `Page` and `RelationObject` and the next value is based on the value form `Page`.`Sort` instead of `RelationObject`.`Sort`.

Renaming the `RelationObject`.`Sort` to something else like `RelationObject`.`MMTSort` will restore expected behaviour.

As far as I can see, this issue is not breaking any functionality. Even if the values are incorrect, they can still be used for sorting.

The concern is that two separate columns are getting mixed up and I don't think relying on unique sort column names is a good idea. I recommend further investigation.

One option would be to not use the MMT list, but using `RelationObject` `DataList` instead when working with sort values.

## Pull requests

- [x] https://github.com/symbiote/silverstripe-gridfieldextensions/pull/272

robbieaverill:
This fixes an error where reordering a many many through list would increment the sort order from the highest value in SiteTree (when `Sort` is the sort order on the MMTL relation model).

The error doesn't actually cause any problems with sorting, but the values it saves are not 1 based but come from the highest value in SiteTree.

This patch re-uses some existing logic to lookup the sort orders from the MMTL relationship table and use those instead, and centralises the logic in a way that both parts of the code can use it.

Fixes #268

---

Side note: this PR also fixes some broken doc blocks.

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -4,6 +4,8 @@ namespace Symbiote\GridFieldExtensions;
 
 use Exception;
 use SilverStripe\Control\Controller;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Forms\GridField\GridField;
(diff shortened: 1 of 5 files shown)


robbieaverill:
Ok that's weird. I couldn't reproduce the postgres test failures locally, and I've manually validated that the code works on postgres. I've bumped the postgres version up to SS 4.2 with pg 2.1.x-dev (which is where we have other supported modules running so it supports transactions) and it's passing now.

src/GridFieldOrderableRows.php
```diff
@@ -759,4 +768,30 @@ protected function getManyManyInspector($list)
         }
         return $inspector;
     }
+
+    /**
+     * Used to get sort orders from a many many through list relationship record, rather than the current
+     * record itself.
+     *
+     * @param ManyManyList|ManyManyThroughList $list
+     * @return int[] Sort orders for the
```
NightJar: Sort orders for the ?

PR #273, opened 2018-10-09
FIX Type hint SS_List instead of DataList and use filter() over where()

robbieaverill:
This allows people to use this component with ArrayList instances if they want to. Protected methods are generally not considered part of our API, which is why I am proposing this change as a patch.

Open to suggestions othewise =)

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -664,12 +664,12 @@ class GridFieldOrderableRows extends RequestHandler implements
         $this->extend('onAfterReorderItems', $list, $values, $sortedIDs);
     }
 
-    protected function populateSortValues(DataList $list)
+    protected function populateSortValues(SS_List $list)
     {
         $list   = clone $list;
         $field  = $this->getSortField();
         $table  = $this->getSortTable($list);
-        $clause = sprintf('"%s"."%s" = 0', $table, $this->getSortField());
+        $clause = [sprintf('"%s"."%s"', $table, $this->getSortField()) =>  0];
         $now = DBDatetime::now()->Rfc2822();
         $additionalSQL = '';
         $baseTable = DataObject::getSchema()->baseDataTable($list->dataClass());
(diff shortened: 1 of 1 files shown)


unclecheese:
I think if the goal is to make this component `ArrayList` compatible, it should also address other `DataList` specific calls, such as the reliance on `->dataClass()`. Or would using an `ArrayList` implicitly mandate that a developer stub out that method in some `ArrayList` subclass?

robbieaverill:
Yeah I'm not really sure. I guess this was more of a quick fix rather than an all encompassing solution. It doesn't make much sense for an ArrayList to have a dataClass, but equally (as you say) extending ArrayList and stubbing out that method wouldn't be a great dev flow either... suggestions?

unclecheese:
Yeah, it doesn't seem like there's an ideal option, here. My fix focused on a subclass of the component that was project-specific. I'll share the code with you outside of this thread.

robbieaverill:
That could be a safer option for sure. I thought about doing that here too, but the type hinting would've prevented us from overloading these methods (easily anyway)

unclecheese:
> So it would only affect you if you've written specific code (e.g. subclasses of this component) which rely on DataList features

I'm not sure that's true.

```php
class MySorter extends GridFieldOrderableRows
{
  protected function populateSortValues(DataList $list);
}
```

It doesn't matter what I'm doing in that class. The definition of the method alone (type hint mismatch) is enough to crash my project on upgrade.

chillu:
Right, good point. And there *are* a few open source subclasses of this actually: https://github.com/search?q=%22extends+GridFieldOrderableRows%22&type=Code

robbieaverill:
Yeah it’s teally only the method signature that is risky, but as noted we don’t consider protected methods part of our API (right?)

unclecheese:
We ended up solving this in a different way, but I think we should leave this open, as having full ArrayList support for this module would be a good long-term improvement. Keen to hear from the Symbiote team on this one.

robbieaverill:
Closing due to lack of feedback. If we're going to implement full SS_List support in this module it'll probably need more than just this pull request, but it'd probably come up when someone searches for it in this repo's GitHub UI.

PR #283, opened 2019-01-22
FIX: Remove unnecessary "version mismatch" restriction (fixes #282)

### LINKED ISSUE #282: "ManyManyThrough cannot mismatch Versioning between join class and related class"
https://github.com/symbiote/silverstripe-gridfieldextensions/blob/3.2.1/src/GridFieldOrderableRows.php#L604-L609, introduced in https://github.com/symbiote/silverstripe-gridfieldextensions/pull/260

I don’t see any reason for this exception to exist - if we’re dealing with an MMTL, we’re updating the sort on the join record and not touching the related record at all. Whether the related record is versioned or not doesn’t matter here, as we’re not inspecting or updating it. If I remove the exception, everything works as expected.

cc @NightJar, do you remember why this was added? 😁

kinglozzer:
Fixes https://github.com/symbiote/silverstripe-gridfieldextensions/issues/282

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -588,25 +588,11 @@ class GridFieldOrderableRows extends RequestHandler implements
         // Model has sort column and is versioned - handle as versioned
         // Model has sort column and is NOT versioned - handle as NOT versioned
         // Model doesn't have sort column because sort column is on ManyManyList - handle as NOT versioned
-        // Model doesn't have sort column because sort column is on ManyManyThroughList...
-        //   - Related item is not versioned:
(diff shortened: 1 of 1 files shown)


robbieaverill:
We can fix those test failures by either using recipe-cms instead of recipe-core, or by locking a specific versioned constraint - in the past I've just changed the recipe =)

robbieaverill:
@NightJar fancy following up on this?

robbieaverill:
I'm happy to merge this. @nightjar feel free to open follow up issues if you find anything you don't like as a result - thanks @kinglozzer!

robbieaverill:
Also, nice branch name

PR #284, opened 2019-01-25
Publish versioned objects on sort

andrewhoule:
When sorting versioned objects the sort value gets saved in the draft table, but not the _live table. This adds publishing for versioned objects.

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -648,6 +648,7 @@ class GridFieldOrderableRows extends RequestHandler implements
                 if ($record->$sortField != $newSortValue) {
                     $record->$sortField = $newSortValue;
                     $record->write();
+                    $record->doPublish();
                 }
             }
         }


robbieaverill:
Hi @andrewhoule, thanks for taking the time to make a pull request for this!

>When sorting versioned objects the sort value gets saved in the draft table, but not the _live table

To me this sounds like expected behaviour. I'd expect to be able to change the order of some records in draft mode before I choose to publish them. I don't think we should make it automatic.

Feel free to reopen this if you disagree, or I've misunderstood something =)

ScopeyNZ:
You're right that some parts of the CMS publish all objects that were affected by another objects re-order. SiteTree for example. This has it's own issues though. Right now there's not really a "good" solution. You might like to follow this issue:

https://github.com/silverstripe/silverstripe-versioned/issues/180

andrewhoule:
Interesting stuff... Thanks for the reference! In the context of what I usually need... Having a client publish the sort silently is a much easier solution than trying to explain versioning and that they would need to publish each item directly. I guess I could try an extension for my use cases. Thanks again for the clarification on all this. 👏🏻

micmania1:
@andrewhoule  #289 added this in but made it configurable. It will only publish something if it was already the latest version and published otherwise it will continue to save as draft.

andrewhoule:
@micmania1 just tested things out... Works great! Thanks!!!

PR #289, opened 2019-03-20
NEW added the ability to autopublish items that are already live

micmania1:


docs/en/index.md
```diff
@@ -114,6 +114,19 @@ class Item extends DataObject {
...
 ```
 
+### Versioning
+By default `GridFieldOrderableRows` will handle versioning but won't automatically publish any records. The user will need to go into each record and publish them manually which could get cumbersome for large lists.
+
+You can configure the list to automatically publish a record if the record is the latest version and is already published. This won't publish any records which have draft changes.
+
+```php
+$orderable = new GridFieldOrderableRows()
+    ->setRebpublishLiveRecords(true);
```
robbieaverill:
```suggestion
    ->setRepublishLiveRecords(true);
```

src/GridFieldOrderableRows.php
```diff
@@ -138,6 +148,28 @@ public function setImmediateUpdate($bool)
...
+    public function getRepublishLiveRecords()
+    {
+        return $this->republishLiveRecords;
+    }
+
+    /**
+     * @see $republishLiveRecords
+     *
+     * @param boolean $bool
+     * @return GridFieldOrderableRows $this
```
robbieaverill:
```suggestion
     * @return $this
```

src/GridFieldOrderableRows.php
```diff
@@ -656,7 +688,13 @@ protected function reorderItems($list, array $values, array $sortedIDs)
...
                 if ($record->$sortField != $newSortValue) {
                     $record->$sortField = $newSortValue;
-                    $record->write();
+
+                    if ($this->getRepublishLiveRecords() && $record->isLiveVersion()) {
+                        $record->write();
+                        $record->copyVersionToStage(Versioned::DRAFT, Versioned::LIVE, true);
+                    } else {
+                        $record->write();
+                    }
```
robbieaverill: You could probably put the `$record->write()` before the condition to avoid writing it twice

ScopeyNZ: You might have to do `$wasLive = $record->isLiveVersion()` before the write though.

src/GridFieldOrderableRows.php
micmania1: This was just for consistency with the method above it

ScopeyNZ:
Vaguely related to the over-arching versioned sorting issue: https://github.com/silverstripe/silverstripe-versioned/issues/180

src/GridFieldOrderableRows.php
micmania1: It only writes once? I can't write before the check since then the check for `isLiveVersion()` would fail.

docs/en/index.md
micmania1: updated

src/GridFieldOrderableRows.php
micmania1: @ScopeyNZ updated

robbieaverill (approved):

purplespider:
Doesn't look like this has made it into recent releases as it was merged into `master` rather than `3`?

Any chance in publishing a release including this feature?

Thanks!

n8-dev:
Just found this via googling and looking.
Would love to get this included.

PR #292, opened 2019-05-12
NEW Auto-publish objects on sort

amolswnz:
Related SS4 PR: #289

### DIFF
--- a/code/GridFieldOrderableRows.php
+++ b/code/GridFieldOrderableRows.php
@@ -463,6 +463,9 @@ class GridFieldOrderableRows extends RequestHandler implements
 		if ($class == $this->getSortTable($list)) {
 			$isVersioned = $class::has_extension('Versioned');
 		}
+		if ($class::has_extension('Versioned')) {
+			$isVersioned = true;
+		}
 
 		// Loop through each item, and update the sort values which do not
 		// match to order the objects.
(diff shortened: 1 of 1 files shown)


code/GridFieldOrderableRows.php
```diff
@@ -463,6 +463,9 @@ protected function reorderItems($list, array $values, array $sortedIDs) {
 		if ($class == $this->getSortTable($list)) {
 			$isVersioned = $class::has_extension('Versioned');
 		}
+		if ($class::has_extension('Versioned')) {
+			$isVersioned = true;
```
ScopeyNZ: I assume you added this to use it later but it's unused?

code/GridFieldOrderableRows.php
```diff
@@ -505,11 +508,15 @@ protected function reorderItems($list, array $values, array $sortedIDs) {
 			// similar to the SiteTree where you change the position, and then
 			// you go into the record and publish it.
 			foreach($sortedIDs as $sortValue => $id) {
-				if($map[$id] != $sortValue) {
-					$record = $class::get()->byID($id);
+				$record = $class::get()->byID($id);
```
ScopeyNZ: It seems like this could be left inside the condition below? It's a sliiiight performance impact moving it out of the condition because you're loading a DataObject for every item regardless of the following condition...

code/GridFieldOrderableRows.php
```diff
@@ -505,11 +508,15 @@ protected function reorderItems($list, array $values, array $sortedIDs) {
...
+				if ($map[$id] != $sortValue) {
 					$record->$sortField = $sortValue;
 					$record->write();
 				}
+				// Update stage to live only for published records
+				if ($record->isPublished()) {
```
ScopeyNZ: This is only relevant if there's a write, so you can move it into the condition. Also this is probably where you want to use `$isVersioned` you prepared earlier.

amolswnz:
I agree, I will change my code.

amolswnz:
Closing this pr, as this issue does not exists in fresh SS3 installation.

PR #294, opened 2019-08-26
FIX: Let GridFieldEditableColumns edit relations via dot syntax

sminnee:
This is a companion to https://github.com/silverstripe/silverstripe-framework/pull/9192 to provide the same functionality for inline editing in GridFields.

A valuable use of this is editing fields in the join-object of a many-many-through relation.

### DIFF
--- a/src/GridFieldAddNewInlineButton.php
+++ b/src/GridFieldAddNewInlineButton.php
@@ -13,6 +13,7 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectInterface;
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\ManyManyList;
+use SilverStripe\ORM\ManyManyThroughList;
 use SilverStripe\View\ArrayData;
 use SilverStripe\View\Requirements;
 use Exception;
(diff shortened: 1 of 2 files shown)


michalkleiner:
Was this change still valid in its original form/shape since last year as there's been some reverting on the `false,false,false,true` argument set for the write method done recently in the framework IIRC. Can you please confirm @sminnee?

sminnee:
@michalkleiner When this change was applied to _all_ GridFieldDetailForms, it caused some edge-case issues. GFDF is used in so many places it ended up being a bit too dangerous and I'm going to reintroduce it as an optional config. The scope of GridFieldEditableColumns is less - i.e. it's used in fewer places - but it it might be worth submitting a further PR so that the change to write() is only activated with a `setWriteRelations()` config param on the gridfield component.

I'm happy to take @nyeholt's guidance on this.

ntd:
@sminnee For some reason, `GridFieldEditableColumns::handleSave` (and for similarity `GridFieldAddNewInlineButton::handleSave`, I suppose)  is not able to save on `ManyManyThroughList` (this is not a regression: it has never been able to do so). In #311 I have a sample code highlighting this issue.

If you have any hint to give it would be really appreciated: trying to come up with a solution, my code is quickly becoming too hackish.

sminnee:
It's possible that you also need this unmerged change, @ntd, which I pull into my project via composer-patches https://github.com/silverstripe/silverstripe-framework/pull/9192

I can't recall if it was only needed for detail forms or for editable gridfields too, sorry. But give it a go, maybe?

PR #302, opened 2020-01-29
Fix support for dot notation in GridFieldOrderableRows

lpostiglione:
When you want to have a GridField for a sub-relation on a relation of your DataObject you use dot notation, this breaks the functionality of GridFieldOrderableRows because in the POST request the dot is replaced by an underscore.


lpostiglione:
Could you integrate this please?

nyeholt:
@lpostiglione apologies for the delay, have merged it in

PR #305, opened 2020-03-17
FIX: Allow silverstirpe/tagfield in GridFieldEditableColumns

sminnee:
This change was necessary to get a silverstripe/tagfield to work inside a GridFieldEditableColumns list. The value was previously loaded correctly, and subsequently clobbered by this line.

From what I can see, the removal of this line doesn’t break anything
else, as the record content is loaded earlier in the call stack by
$form->loadDataFrom() in getForm().

### DIFF
--- a/src/GridFieldEditableColumns.php
+++ b/src/GridFieldEditableColumns.php
@@ -97,7 +97,6 @@ class GridFieldEditableColumns extends GridFieldDataColumns implements
         $value = $this->formatValue($grid, $record, $col, $value);
 
         $field->setName($this->getFieldName($field->getName(), $grid, $record));
-        $field->setValue($value);
 
         if ($field instanceof HtmlEditorField) {
             return $field->FieldHolder();


dhensby (approved):

sminnee:
@robbieaverill or anyone; any thoughts?

src/GridFieldEditableColumns.php
```diff
@@ -97,7 +97,6 @@ public function getColumnContent($grid, $record, $col)
         $value = $this->formatValue($grid, $record, $col, $value);
```
robbieaverill, line 97: Am I right in thinking that now this line and the previous assignments of `$value` are obsolete?

chillu: Correct, `$value` isn't used in the rest of the function and should be removed

chillu:
I've tested this out, and I can't reproduce how this broke - can you recall the error @sminnee? Here's my test scenario, which covers field formatting. This works without the change in this PR (add new items, add/remove tags)

```
composer create-project silverstripe/installer:4 4
cd 4/
composer require silverstripe/tagfield silverstripe/taxonomy symbiote/silverstripe-gridfieldextensions
```

```php
<?php

use SilverStripe\ORM\FieldType\DBInt;
use SilverStripe\ORM\FieldType\DBCurrency;
use SilverStripe\ORM\FieldType\DBText;
use SilverStripe\Taxonomy\TaxonomyTerm;

class TestItem extends DataObject
{
    private static $db = [
        'MyText' => DBText::class,
        'MyInt' => DBInt::class,
        'MyCurrency' => DBCurrency::class,
    ];

    private static $many_many = [
        'Tags' => TaxonomyTerm::class,
    ];
}
```

```php
<?php

namespace {

    use SilverStripe\CMS\Model\SiteTree;
    use SilverStripe\Forms\GridField\GridFieldButtonRow;
    use SilverStripe\Forms\GridField\GridFieldConfig;
    use SilverStripe\Forms\GridField\GridFieldDeleteAction;
    use SilverStripe\Forms\GridField\GridFieldToolbarHeader;
    use SilverStripe\Forms\TextField;
    use SilverStripe\Forms\ReadonlyField;
    use SilverStripe\TagField\TagField;
    use SilverStripe\Taxonomy\TaxonomyTerm;
    use Symbiote\GridFieldExtensions\GridFieldAddNewInlineButton;
    use Symbiote\GridFieldExtensions\GridFieldTitleHeader;

    class Page extends SiteTree
    {
        public function getCMSFields()
        {
            $fields = parent::getCMSFields();

            $grid = new GridField(
                'Test',
                'Test',
                TestItem::get(),
                GridFieldConfig::create()
                    ->addComponent(new GridFieldButtonRow('before'))
                    ->addComponent(new GridFieldToolbarHeader())
                    ->addComponent(new GridFieldTitleHeader())
                    ->addComponent(new GridFieldEditableColumns())
                    ->addComponent(new GridFieldDeleteAction())
                    ->addComponent(new GridFieldAddNewInlineButton())
            );
            $grid->getConfig()->getComponentByType(GridFieldEditableColumns::class)->setDisplayFields([
                'MyText'  => function($record, $column, $grid) {
                    return new TextField($column);
                },
                'MyInt' => [
                    'field' => ReadonlyField::class
                ],
                'MyCurrency' => true,
                'Tags' => function($record, $column, $grid) {
                    return (new TagField('Tags', 'Tags', TaxonomyTerm::get(), $record->Tags()))
                        ->setTitleField('Name');
                }
                ]);


            $fields->addFieldToTab('Root.TestItems', $grid);

            return $fields;
        }
    }
}
```

I'm a bit worried about unintended side effects, especially not re-invoking field formatting.

chillu:
Recommend that we close this without merging. No feedback in months, I can't reproduce, and I think it's too risky. I don't have permissions to close this myself though. Do any maintainers agree?

@sminnee We're also not using inlined tagfields in the internal RMS any more, so I'm not worried about breaking anything there either.

PR #313, opened 2020-07-29
FIX Reinstate previous field fetch logic as fallback

Cheddam:
Fixes some cases where a field is not retrievable via `dataFieldByName()`.

Resolves [silverstripe/silverstripe-userforms#994](https://github.com/silverstripe/silverstripe-userforms/issues/994).

### DIFF
--- a/src/GridFieldEditableColumns.php
+++ b/src/GridFieldEditableColumns.php
@@ -84,6 +84,12 @@ class GridFieldEditableColumns extends GridFieldDataColumns implements
             $value  = $grid->getDataFieldValue($record, $col);
             $field = $fields->dataFieldByName($col);
 
+            // Fall back to previous logic
+            if (!$field) {
+                $rel = (strpos($col, '.') === false); // field references a relation value
+                $field = ($rel) ? clone $fields->fieldByName($col) : new ReadonlyField($col);
+            }
+
             if (!$field) {
                 throw new Exception("Could not find the field '$col'");
             }


emteknetnz:
@Cheddam What's unit test coverage on this like?

Cheddam:
@emteknetnz Zero, and unfortunately it looks non-trivial to introduce any. A more sure-fire fix would be to simply revert the code change entirely, but I've manually tested this against the module it has impacted most severely (Userforms) and it restores operation as expected.

brynwhyman:
+1 to moving forward as opposed to reverting previous commits (which are over a month old).

emteknetnz:
PR that changed logic https://github.com/symbiote/silverstripe-gridfieldextensions/pull/294/files#diff-086e318c40682b987d17294b79b8d741L85

emteknetnz:
@Cheddam So, will implementing this fallback means there's a chance something that was supposed to be fixed in the PR that introduced regression (PR linked above) may now no longer work (because it can run `fieldByName` instead of only`dataFieldByName`) ?

Cheddam:
Not unless the expected outcome was an exception? That's why this PR reinstates the old logic as a _fallback_ if the new logic fails to fetch the field, rather than the other way around.

emteknetnz (approved):
I'm OK to approve as is

Given you've just re-implemented old code, and not written new code yourself, and you've tested it locally, I think it's OK for you to self-merge at this point

If you think this requires further testing, then shout out.  Otherwise, I think it's OK for you to merge as is now.

PR #314, opened 2020-08-07
FIX: Fixed handling of uneditable records in GridFieldEditableColumns

### LINKED ISSUE #312: GridFieldEditableColumns with callbacks hides column values on records without edit permissions
When using callbacks in `setDisplayFields()`, they work differently than in the parent implementation: It expects a `FormField` instance, rather than a scalar value. `GridFieldEditableColumns ->getColumnContent()` defers to the parent implementation when a record isn't editable, which makes sense. But that parent implementation also evaluates the callback in `setDisplayFields()`, and doesn't know what to do with a form field instance.

```php
<?php
(new GridFieldEditableColumns())
    ->setDisplayFields([
        'Comment' => ['callback' => function ($record, $column, $grid) {
            // return some form field, not a scalar value
        }]);
```

 * Option A: Name the `callback` differently (e.g. `callback_formfields`) to avoid the confusion about the different return values. You could implement this in a backwards compatible way by falling back to `callback`
 * Option B: Don't defer to parent implementation, but rather go through the form field creation logic, and then call `performReadonlyTransformation()` on it before returning

sminnee:
Fixes #312

To do:

 * [x] add test
 * [x] Sam to check that a bug he found in his project isn't due to this fix being inadequate

### DIFF
--- a/src/GridFieldEditableColumns.php
+++ b/src/GridFieldEditableColumns.php
@@ -100,7 +100,7 @@ class GridFieldEditableColumns extends GridFieldDataColumns implements
         $field->setName($this->getFieldName($field->getName(), $grid, $record));
         $field->setValue($value);
 
-        if (!$record->canEdit()) {
+        if ($grid->isReadonly() || !$record->canEdit()) {
             $field = $field->performReadonlyTransformation();
         }
 
--- a/src/GridFieldEditableColumns.php
+++ b/src/GridFieldEditableColumns.php
@@ -56,10 +56,6 @@ class GridFieldEditableColumns extends GridFieldDataColumns implements
 
     public function getColumnContent($grid, $record, $col)
     {
-        if (!$record->canEdit()) {
-            return parent::getColumnContent($grid, $record, $col);
-        }
-
         $fields = $this->getForm($grid, $record)->Fields();
 
         if (!$this->displayFields) {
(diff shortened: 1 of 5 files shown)


robbieaverill (approved):
Looks reasonable to me, I haven't tested it out though

sminnee:
I guess it could use a test eh ;-)

sminnee:
I noticed that a project I'm working on had a similar issue on spite of this fix so I'm going to double-check that this is working before merge.

**EDIT:** The issue that I ran into was that a readonly transformation removes GridFieldEditableColumns. The GridField API doesn't provide any mechanism for this on a default/application-wide basis, so the fix is to call this on every grid field that uses GridFieldEditableColumns:

```php
$gridField->setReadonlyComponents(array_merge($gridField->getReadonlyComponents(), [ GridFieldEditableColumns::class ]));
```

This works if a grid field has been made readonly because canEdit() permissions are missing. However, if a gridfield is optionally made readonly, it won't work. So c0e698a has been added to address this.

src/GridFieldEditableColumns.php
```diff
@@ -104,6 +100,10 @@ public function getColumnContent($grid, $record, $col)
         $field->setName($this->getFieldName($field->getName(), $grid, $record));
         $field->setValue($value);
 
+        if (!$record->canEdit() || $grid->isReadonly()) {
```
dhensby: tiny optimisation may bee to check `isReadOnly` first as I assume that's a much less-intensive operation in the case that users can edit (typical) and that it is read only?

dhensby (approved):

src/GridFieldEditableColumns.php
sminnee: Fair enough - force-pushing a fix.

dhensby (approved):

robbieaverill (approved):
LGTM

PR #321, opened 2021-06-03
Add translation string for Toast message "Records reordered."

TomOudeRengerink:
Added a translation string for the toast message "Records reordered.".

This message is shown when the GridFieldOrableRows "handReorder" method is called.

I have provided translations for all of the available language files. Please note that these words are translated through Google Translate and therefore might not be 100% correct.

## Issues
- https://github.com/symbiote/silverstripe-gridfieldextensions/issues/378

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -456,7 +456,10 @@ class GridFieldOrderableRows extends RequestHandler implements
             $this->httpError(400);
         }
 
-        Controller::curr()->getResponse()->addHeader('X-Status', rawurlencode('Records reordered.'));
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Thank you for this contribution and sorry it's taken so long to review.

I've removed the translations - please see [our translation contribution guide](https://docs.silverstripe.org/en/5/contributing/translations/) to learn how to contribute those.

I've retargetted and rebased this PR against the `3.6` branch and updated the localisation key to be based on the class name, which is consistent with best practices.

emteknetnz (approved):
Seems good

`rawlurlencode()` looks like it's required because of the use of `decodeURIComponent()` in js files

Issue #324, opened 2021-08-24
GridFieldOrderableRows does not support secondary or default sort fields.


davejtoews:
I frequently have orderable objects which should have a default sort order, but option for manually overriding. So the object will have something like:

```php
private static $default_sort = 'Sort ASC, Title ASC`;
```

With the `GridFieldOrderableRows` component added, only the sort column managed by the component itself is considered in the CMS display of the items. If no values have been set in the sort field they display unsorted in whichever order SQL decides to show them. (Mostly just ordered by id).

The component offers a getter and setter for `$extraSortFields` but these are applied as higher priority sorts than the managed sort column. What I'm looking for is setting sorts with a lower priority than the managed sort column, preferably making use of the defaults set elsewhere.

PRs
- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/325
- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/329
- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/349

GuySartorelli:
Closing as a duplicate of https://github.com/symbiote/silverstripe-gridfieldextensions/issues/357 (we took the latter through backlog refinement today)

PR #325, opened 2021-08-25
🐛 (cms interface) applies default sort order to CMS ordering

davejtoews:
This allows secondary sort ordering to be applied when it exists and when the objects have not been sorted manually.

This attempts to address #324. It could use some sanity checks and testing, but it is working as needed in my application. It was an open question to me whether this was the best way to address it or whether additional sort fields should be applied via a similar mechanism as what exists for `$extraSortFields`. My preference was for the component to automatically pick up on default sorts, but applying sort through another method would make sense to me too.

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -364,11 +364,10 @@ class GridFieldOrderableRows extends RequestHandler implements
 
                 if ($list instanceof DataList) {
                     $classname = $list->dataClass();
-                    if (Config::inst()->get($classname, 'default_sort')) {
+                    if ($defaultSort = Config::inst()->get($classname, 'default_sort')) {
                         // Append the default sort to the end of the sort string
                         // This may result in redundancy... but it seems to work
-                        $sortterm .= $sortterm ? ', ' : '';
-                        $sortterm .= Config::inst()->get($classname, 'default_sort');
+                        $sortterm .= ($sortterm ? ', ' : '') . $defaultSort;
                     }
                 }
             }
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -8,6 +8,7 @@ use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Config\Config;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\GridField\GridField_ColumnProvider;
 use SilverStripe\Forms\GridField\GridField_DataManipulator;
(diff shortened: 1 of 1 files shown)


src/GridFieldOrderableRows.php
```diff
@@ -359,7 +360,15 @@ public function getManipulatedData(GridField $grid, SS_List $list)
                 // Fix bug in 3.1.3+ where ArrayList doesn't account for quotes
                 $sortterm .= $this->getSortTable($list).'.'.$this->getSortField();
             } else {
+                // If not an ArrayList it's definitely a DataList right?
```
michalkleiner: Nope. The param says `SS_List` (which is an interface) and there are more classes implementing it than just `ArrayList` and `DataList`. It's also extended to `Sortable` which may be implemented by yet another set of classes, so we can't assume the `->dataClass()` method is available without further checks.

davejtoews: You're right! I missed the sub-interfaces of SS_List.

Happy to alter the PR. I'm thinking of just directly checking `method_exists($list, 'dataClass')` or do you think `$list instanceof DataList` would be cleaner?

michalkleiner: I would check for the `DataList` and only apply the default sort as the secondary ordering there

src/GridFieldOrderableRows.php
```diff
@@ -359,7 +360,15 @@ public function getManipulatedData(GridField $grid, SS_List $list)
...
+                // Or do we need another conditional here?
                 $sortterm .= '"'.$this->getSortTable($list).'"."'.$this->getSortField().'"';
+                $classname = $list->dataClass(); // DataLists have a dataClass
+                if (Config::inst()->get($list->dataClass(), 'default_sort')) {
+                    // Append the default sort to the end of the sort string
+                    // This may result in redundancy... but it seems to work
+                    $sortterm .= ', ' . Config::inst()->get($classname, 'default_sort');
```
michalkleiner: `$sorttterm` can be an empty string here, so we can't start with the comma here, will need further logic.

src/GridFieldOrderableRows.php
davejtoews: Checking for a DataList now.

src/GridFieldOrderableRows.php
davejtoews: Adjusted to only add a comma  when the string is not empty.

src/GridFieldOrderableRows.php
```diff
@@ -360,7 +361,18 @@ public function getManipulatedData(GridField $grid, SS_List $list)
...
             } else {
                 $sortterm .= '"'.$this->getSortTable($list).'"."'.$this->getSortField().'"';
+
+                if ($list instanceof DataList) {
+                    $classname = $list->dataClass();
+                    if (Config::inst()->get($classname, 'default_sort')) {
+                        // Append the default sort to the end of the sort string
+                        // This may result in redundancy... but it seems to work
+                        $sortterm .= $sortterm ? ', ' : '';
+                        $sortterm .= Config::inst()->get($classname, 'default_sort');
```
michalkleiner:
```suggestion
                    if ($defaultSort = Config::inst()->get($classname, 'default_sort')) {
                        // Append the default sort to the end of the sort string
                        // This may result in redundancy... but it seems to work
                        $sortterm .= ($sortterm ? ', ' : '') . $defaultSort;
```

emteknetnz (approved):

emteknetnz:
I've created a follow up unit-test PR for this here https://github.com/symbiote/silverstripe-gridfieldextensions/pull/329

PR #329, opened 2022-02-02
MNT Add unit tests for default sort

emteknetnz:
Issue https://github.com/symbiote/silverstripe-gridfieldextensions/issues/324

Unit tests for https://github.com/symbiote/silverstripe-gridfieldextensions/pull/325

### DIFF
--- a/tests/GridFieldOrderableRowsTest.php
+++ b/tests/GridFieldOrderableRowsTest.php
@@ -6,6 +6,7 @@ use ReflectionMethod;
 use SilverStripe\Dev\SapphireTest;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\GridField\GridFieldConfig_RelationEditor;
+use SilverStripe\ORM\DataList;
 use Symbiote\GridFieldExtensions\GridFieldOrderableRows;
 use Symbiote\GridFieldExtensions\Tests\Stub\PolymorphM2MChild;
 use Symbiote\GridFieldExtensions\Tests\Stub\PolymorphM2MMapper;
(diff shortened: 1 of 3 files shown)


tests/GridFieldOrderableRowsTest.php
```diff
@@ -296,4 +301,41 @@ public function testReorderItemsSubclassVersioned()
...
+        ]);
+        $this->assertSame(['A', 'B', 'C'], $sortedList->column('Title'));
+    }
+
+    public function testGetManipulatedDataWithDefaultSort()
+    {
+        $sortedList = $this->getTitleSortedListForManipuatedData(TitleSortedObject::class, [
+            ['Title' => 'Z', 'Iden' => 'C', 'DefaultSort' => 3],
+            ['Title' => 'Z', 'Iden' => 'A', 'DefaultSort' => 2],
+            ['Title' => 'Z', 'Iden' => 'B', 'DefaultSort' => 1],
```
ChloeSartorelli, line 320: Why are we using `Iden` here instead of `Title`? It seems like it would be a lot clearer if we use `Title` like with the above test, so we can clearly contrast the results of _this_ sorting vs the sorting by title that happens above?

emteknetnz: This is an old PR so I cannot remember exact reasoning

Best guess is that in `getTitleSortedListForManipuatedData()` we have `$orderable = new GridFieldOrderableRows('Title');`, and I wanted to test what happens with `Title` is the same value, so it falls back to `$default_sort`

ChloeSartorelli: Fair enough.

GuySartorelli (approved):

PR #336, opened 2022-03-03
ENH: Update reorderItems() to use ORM where possible

### LINKED ISSUE #335: Question: Could GridFieldOrderableRows::reorderItems() please use ORM for non-Versioned models?
## Current implementation

When a model does not implement the `Versioned` extension, `reorderItems()` uses SQL queries to update the sort values of models.

When a model **does** implement the `Versioned` extension, `reorderItems()` sets the sort value on the record and performs a `write()` through the ORM.

## Problem statement

When developers implement `onBeforeWrite()` and `onAfterWrite()` methods, there is some confusion on whether or not these methods will be called when the records are reordered. If they are `Versioned`, then these methods will be triggered; if they are not `Versioned`, then they will not.

## Question

Could we please standarise the `reorderItems()` method to always use the ORM to save records?

## Considerations

Would this sort of change need to be a new major version, or would it be considered a "bugfix"?

chrispenny:
## Description

Closes #335

Based onto `3`, but please feel free to update that with what you think is appropriate.

I'm not 100% familiar with all of the different `Lists` that we have, but **I think** it would only be the `ManyManyList` that we still need to process with SQL Queries, as their DB records have no representation in the ORM.

## Additions

Just added a cheeky little extension point for `onBeforeReorderItems`, because "why not?".

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -587,9 +587,10 @@ class GridFieldOrderableRows extends RequestHandler implements
      */
     protected function reorderItems($list, array $values, array $sortedIDs)
     {
+        $this->extend('onBeforeReorderItems', $list, $values, $sortedIDs);
+
         // setup
         $sortField = $this->getSortField();
-        $class = $list->dataClass();
         // The problem is that $sortedIDs is a list of the _related_ item IDs, which causes trouble
         // with ManyManyThrough, where we need the ID of the _join_ item in order to set the value.
         $itemToSortReference = ($list instanceof ManyManyThroughList) ? 'getJoin' : 'Me';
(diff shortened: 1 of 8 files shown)


michalkleiner:
I'm all for simplification, though it seems suspicious that such a big removal will go without an impact in tests. Does it mean something significantly changed somewhere else to still support the reordering?

chrispenny:
Hi @michalkleiner,

The module already supported sorting through the ORM, the only thing was that it only used the ORM for `Versioned` DataObjects. All I've done is change the criteria for when the module uses the ORM vs SQL Queries.

The relevant code for sorting through the ORM is still here:
https://github.com/symbiote/silverstripe-gridfieldextensions/pull/336/files#diff-ad5bbd1d14499e77c6ef3223c6c3746fa71333fec5831b4c3fcfa7a5e40f09a0R618

chrispenny:
Hi @michalkleiner 

We now have tests covering:

* `has_many` un-`Versioned`
* `has_many` subclass un-`Versioned`
* `has_many` subclass `Versioned`
* `many_many` un-`Versioned`
* `many_many` `Versioned`
* `many_many` with through model un-`Versioned`
* `many_many` with through model `Versioned`

Please let me know if there is anything else I can do to increase confidence in this change.

src/GridFieldOrderableRows.php
```diff
@@ -587,53 +588,20 @@ protected function reorderItems($list, array $values, array $sortedIDs)
...
-            $class = $this->getManyManyInspector($list)->getJoinClass();
-            $isVersioned = $class::create()->hasExtension(Versioned::class);
-        } elseif (!$this->isManyMany($list)) {
-            $isVersioned = $class::create()->hasExtension(Versioned::class);
-        }
-
-        // Loop through each item, and update the sort values which do not
-        // match to order the objects.
-        if (!$isVersioned) {
+        // ManyManyList models are not represented by the ORM, and so they need to be updated through an SQL Query
```
ChloeSartorelli: > ManyManyList models are not represented by the ORM

`ManyManyList` is literally the ORM representation for this relation type... did you mean the relation itself doesn't have a `DataObject` backing it up (i.e. the `sort` is on a table that doesn't have a `DataObject` representation)?

If so, maybe a better way of wording this is "ManyManyList extra fields aren't easily updated via the ORM, and so they need to be updated through an SQL Query"

Also now that I've noticed that this is feature missing in the ORM I'm going to see if I can quickly whip up a PR to add it in. Seems like an obvious thing to want.

GuySartorelli:
I'm slowly reviewing this inbetween other work.
If someone else has more time and wants to review it, please go ahead - but otherwise I'll get there eventually.

src/GridFieldOrderableRows.php
chrispenny: > ManyManyList is literally the ORM representation for this relation type... did you mean the relation itself doesn't have a DataObject backing it up (i.e. the sort is on a table that doesn't have a DataObject representation)?

Yup, that's exactly what I meant.

Thanks for the suggestion 😃 I'll update the comment.

ChloeSartorelli: FYI that PR is up: https://github.com/silverstripe/silverstripe-framework/pull/10406
Now I can get back to reviewing this one 😅

GuySartorelli (approved):
Looks good, works correctly locally, and the new tests are fantastic. Thanks for doing this.
Can you please rebase and resolve the merge conflict?

chrispenny:
Thanks, @GuySartorelli! Rebase complete.

I'm not sure why, but running the workflow needs approval, if you could please 😃 .

PR #339, opened 2022-05-19
Allow publishing of already published versioned objects for 3.3 branch

n8-dev:
Similiar to #340  just for the `3.3` branch

---

The change here  #289 looks like it was merged to `master` by mistake.

This is opened to allow a `3.3.n` a release bump as `3.4.n` has some changes to dependencies that not everyone has due to other constraints


dhensby:
As with the other PR, this can't go into patch releases. If you want a feature back-ported, it's best to request a maintainer does that as PRs are often not the best way to do it.

n8-dev:
Right, lodge an issue then @dhensby ?

PR #332, opened 2022-02-12
NEW Extend new AbstractGridFieldComponent class

GuySartorelli:
This makes this module's `GridFieldComponent` classes `Injectable`, and allows any future enhancements in the new abstract class to automatically apply without requiring additional changes in this module.

The class is introduced in silverstripe/framework 4.11.0 (see silverstripe/silverstripe-framework#10204) so the dependency constraint needs to be updated.

Also update docs to encourage use of dependency injection.

PR #340, opened 2022-05-19
Allow publishing of already published versioned objects for 3.4 branch

n8-dev:
Similiar to #339 just for the `3.4` branch

--- 

The change here https://github.com/symbiote/silverstripe-gridfieldextensions/pull/289 looks like it was merged to master by mistake.

This is opened to allow a 3.4.n a release bump

### DIFF
--- a/docs/en/index.md
+++ b/docs/en/index.md
@@ -57,6 +57,14 @@ $grid->getConfig()->getComponentByType('GridFieldEditableColumns')->setDisplayFi
 Editing data contained in `many_many_extraFields` is supported - just treat it as you would any
 other field.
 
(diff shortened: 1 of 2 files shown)


dhensby:
This can't go into patch release branches as it is not a bug fix but a new feature. At best this should go into the 3 branch

n8-dev:
I do agree @dhensby that its anti semver. but its such a pain it wasn't done when it should have been. :(

Esp with `3` branch requiring `"silverstripe/framework": "^4.11"` at ec93c994f8de123233631715138a7f9fdbc9f13c #332 

IMO That should have been 4 branch because that's a significant breaking change for some projects currently considering how new `4.11` is

PR #341, opened 2022-06-01
ENH: Add performance fixes on saving editable columns

satrun77:
The changes in the PR to address the issue described in https://github.com/silverstripe/silverstripe-userforms/issues/773

This might not eliminate the issue completely, but it is an improvement to the current code.

Changes include,
- Use one query to fetch all items needed to be saved, instead of multiple queries.
- Only save items that are changed.

### DIFF
--- a/src/GridFieldEditableColumns.php
+++ b/src/GridFieldEditableColumns.php
@@ -19,10 +19,10 @@ use SilverStripe\Forms\GridField\GridField_URLHandler;
 use SilverStripe\Forms\HTMLEditor\HTMLEditorField;
 use SilverStripe\Forms\LiteralField;
 use SilverStripe\Forms\ReadonlyField;
+use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectInterface;
-use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\ManyManyList;
 
 /**
@@ -130,14 +130,23 @@ class GridFieldEditableColumns extends GridFieldDataColumns implements
         /** @var GridFieldOrderableRows $sortable */
         $sortable = $grid->getConfig()->getComponentByType(GridFieldOrderableRows::class);
 
+        // Fetch the items before processing them
+        $ids = array_keys($value[self::POST_KEY]);
+        if (empty($ids)) {
+            return;
+        }
+        $itemsCollection = ArrayList::create($list->filter('ID', $ids)->toArray());
+
         foreach ($value[self::POST_KEY] as $id => $fields) {
             if (!is_numeric($id) || !is_array($fields)) {
                 continue;
             }
 
-            $item = $list->byID($id);
+            // Find the item from the fetched collection of items
+            $item = $itemsCollection->find('ID', $id);
 
-            if (!$item || !$item->canEdit()) {
+            // Skip not found item, or don't have any changed fields, or current user can't edit
+            if (!$item || !$this->isChanged($item, $fields) || !$item->canEdit()) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


src/GridFieldEditableColumns.php
```diff
@@ -323,4 +333,20 @@ protected function getFieldName($name, GridField $grid, DataObjectInterface $rec
             $name
         );
     }
+
+    /**
+     * Whether or not an object in the grid field has changed data.
+     *
+     * @return bool
+     */
+    protected function isChanged(DataObject $item, array $fields)
```
emteknetnz:
```suggestion
     * Whether or not an object in the grid field has changed data.
     */
    private function isChanged(DataObject $item, array $fields): bool
```

src/GridFieldEditableColumns.php
```diff
@@ -130,12 +130,22 @@ public function handleSave(GridField $grid, DataObjectInterface $record)
         /** @var GridFieldOrderableRows $sortable */
         $sortable = $grid->getConfig()->getComponentByType(GridFieldOrderableRows::class);
 
+        // Fetch the items before processing them
+        $ids = array_keys($value[self::POST_KEY]);
```
emteknetnz, line 134:
```suggestion
        $ids = array_keys($value[self::POST_KEY]);
        if (empty($ids) {
            return;
        }
```
Filtering by an array of values will throw an exception when the array is empty, limitation of the ORM

src/GridFieldEditableColumns.php
```diff
@@ -130,12 +130,22 @@ public function handleSave(GridField $grid, DataObjectInterface $record)
...
             }
 
-            $item = $list->byID($id);
+            // Find the item from the fetched collection of items
+            $item = $itemsCollection->filter('ID', $id)->first();
+
+            // Skip items that does not have any changed fields
+            if (!$this->isChanged($item, $fields)) {
+                continue;
+            }
```
emteknetnz:
```suggestion
            if (!$item || !$this->isChanged($item, $fields) || !$item-canEdit()) {
                continue;
            }
```
Conbine this with the conditional block below

src/GridFieldEditableColumns.php
```diff
@@ -130,12 +130,22 @@ public function handleSave(GridField $grid, DataObjectInterface $record)
...
+        $itemsCollection = ArrayList::create($list->filter('ID', $ids)->toArray());
+
         foreach ($value[self::POST_KEY] as $id => $fields) {
             if (!is_numeric($id) || !is_array($fields)) {
                 continue;
             }
 
-            $item = $list->byID($id);
+            // Find the item from the fetched collection of items
+            $item = $itemsCollection->filter('ID', $id)->first();
```
emteknetnz:
```suggestion
            $item = $itemsCollection->find('ID', $id);
```

emteknetnz (changes requested):
If we change the protected method to private then there's no API change so we can do a patch release meaning we release it much faster.

Please rebase/retarget to the `3.4` branch

Also prefix your commit message and the title of the PR with `ENH` as this is an enhancement

src/GridFieldEditableColumns.php
```diff
@@ -130,14 +130,23 @@ public function handleSave(GridField $grid, DataObjectInterface $record)
...
                 continue;
             }
 
-            $item = $list->byID($id);
+            // Find the item from the fetched collection of items
+            $item = $itemsCollection->find('ID', $id);
 
-            if (!$item || !$item->canEdit()) {
+            // Skip not found item, or don't have any changed fields, or current user can't edit
+            if (!$item || !$this->isChanged($item, $fields) || !$item-canEdit()) {
```
emteknetnz:
```suggestion
            if (!$item || !$this->isChanged($item, $fields) || !$item->canEdit()) {
```

emteknetnz (changes requested):
Have tested locally - I noticed a syntax error - I fixed on my local and this seemed to work correctly.

Happy to merge after this is fixed

emteknetnz (approved):

emteknetnz:
I've released [3.4.1](https://github.com/symbiote/silverstripe-gridfieldextensions/releases/tag/3.4.1) for you

PR #353, opened 2023-01-16
FIX Use DataList::orderBy()

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10645

### DIFF
--- a/tests/GridFieldOrderableRowsTest.php
+++ b/tests/GridFieldOrderableRowsTest.php
@@ -6,6 +6,7 @@ use ReflectionMethod;
 use SilverStripe\Dev\SapphireTest;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\GridField\GridFieldConfig_RelationEditor;
+use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataList;
 use Symbiote\GridFieldExtensions\GridFieldOrderableRows;
 use Symbiote\GridFieldExtensions\Tests\Stub\PolymorphM2MMapper;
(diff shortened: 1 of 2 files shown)


dhensby:
Shouldn't there be a consistent interface across all lists?

emteknetnz:
SS_List does not specify sort() (or the new orderBy() which is DataList only) - https://github.com/silverstripe/silverstripe-framework/blob/4/src/ORM/SS_List.php

We need to use DataList::orderBy() because it's specifically for using raw SQL.  DataList::sort() no longer accepts raw SQL.  That change was a security enhancement to reduce the surface area vulnerable to SQL injection attacks.

maxime-rainville (changes requested):
Minimally, I would like to have a unit test to cover this scenario. 

The original error says `Uncaught InvalidArgumentException: Invalid sort column "EditableFormField"."Sort"`. Maybe, I'm being naive here, but shouldn't we be able to do something like `FooBar::get()->sort('Relation.SortField')`? Are you meant to be able to use the Dot notation in `sort`?

Side note: I'm not a big fan of how these `$sortterm`. Seems like at least part of this could made to use the regular sort.

emteknetnz:
You can still use dot notation in sort(), in this instance it would just be `->sort('Sort')`, not `->sort('EditableFormField.Sort'`, because we're in the context of an `EditableFormField`.  However in this method there's a bunch of raw SQL being written e.g. `$sortterm .= "$col $dir, ";`, `$sortterm .= ($sortterm ? ', ' : '') . $defaultSort;`, so we need to use `orderBy()` here to retain the current logic.

Have added unit test

tests/GridFieldOrderableRowsTest.php
```diff
@@ -346,6 +330,63 @@ private function getTitleSortedListForManipuatedData(string $dataClass, array $d
         $config->addComponent($orderable);
         $grid = new GridField('MyName', 'MyTitle', $list, $config);
         $sortedList = $orderable->getManipulatedData($grid, $list);
-        return $sortedList;
+        $col = $dataClass === TitleObject::class ? 'Title' : 'Iden';
+        $this->assertSame($expected, $sortedList->column($col));
+    }
+
+    public function provideGetManipulatedData(): array
```
maxime-rainville, line 337: Love the DataProvider. That's much nicer.

maxime-rainville (approved):
Can live with this. Tested that the new unit test fail without the fix. Feel free to Merge on Green.

> However in this method there's a bunch of raw SQL being written e.g. $sortterm .= "$col $dir, ";, $sortterm .= ($sortterm ? ', ' : '') . $defaultSort;

That's what I mean by this could easily be rewritten to work with sort. Maybe something like:
```php
$sortterm[$col] = $dir;
```

emteknetnz:
Note the `4 of 6 checks passed` was github actions being weird, the 'context' and 'generate matrix' had duplicate jobs created which were stuck in yellow.  All the actual jobs were green.

PR #358, opened 2023-03-07
Make link color white.

Sitetools:
There is a regression where the Add new record popup link is now blue. 
This update overwrites the SilverStripe style (/resources/vendor/silverstripe/admin/client/dist/styles/bundle.css)
```css
.ui-widget-content a, .ui-widget.ui-widget-content a, .ui-widget a {
    color: #0071c4;
}
```

Before:
Screenshot from 2023-12-19 13-48-08

After:
Screenshot from 2023-12-19 13-48-17

### DIFF
--- a/css/GridFieldExtensions.css
+++ b/css/GridFieldExtensions.css
@@ -215,7 +215,7 @@
 
 }
 .grid-field-inline-new--multi-class-list a {
-  color: #FFF;
+  color: #FFF !important;
   display: block;
   margin: 5px -10px;
   padding: 0 10px;


GuySartorelli (approved):
Looks good - I was going to ask you to not use `!important` but this css file already uses that in a lot of places, and there's not really a clean way to handle this particular scenario, so if we didn't use `!important` here we'd have to trade one horror for another.

I've retargetted this PR to the `3.6` and reset commits - no change to the actual PR itself.
Will merge when CI goes green.

GuySartorelli:
This will be automatically tagged by GitHub actions. Thanks for contributing!

PR #359, opened 2023-03-08
ENH Update translations

GuySartorelli:
Also add to transifex: https://www.transifex.com/silverstripe/silverstripe-gridfieldextensions/

Note that the `javascript/` and `css/` files not being inside `client/` is unusual, but it's not something we can change in this major release line.
I missed it in the CMS 5 lineup so I'll raise a new card to deal with it for that release line.

## Issue
- https://github.com/silverstripe/.github/issues/12


javascript/GridFieldExtensions.js
```diff
@@ -243,7 +243,11 @@
 				const classes = this.data('classes');
 				const liHtml = Object.keys(classes).map(className => {
 					const link = hrefTemplate.replace('{class}', className);
-					return `<li><a href="${link}">Add: <i>${classes[className]}</i></a></li>`;
+					const linkText = ss.i18n.inject(
```
emteknetnz, line 246: If we're changing this then shouldn't there be an updated bundle.js too?

emteknetnz (changes requested):

javascript/GridFieldExtensions.js
ChloeSartorelli: Why? This is just making strings translatable. It doesn't affect how (or even if) it's transpiled.

emteknetnz: We'll if there was a bundle.js we'd need to update it so that translations were actually used :-)

It's a non issue though

`Requirements::javascript('symbiote/silverstripe-gridfieldextensions:javascript/GridFieldExtensions.js');`

emteknetnz (approved):

PR #366, opened 2023-07-02
Making the module SSv5 compatible

catharsisjelly:
Please see the comments I have added in the actual code change but below is a summary

* Added definitions for class vars
* Converted all long arrays to short
* Changed `composer.json` to use SSv5 stuff
* Fixed bug that appeared from SSv5 change on sorting (see file comment)
* Changed to PSR12 and fixed any CS issues

These changes break backward functionality and will require a major version number increase


src/GridFieldOrderableRows.php
```diff
@@ -384,29 +383,25 @@ public function getManipulatedData(GridField $grid, SS_List $list)
                         $sortterm .= "$col $dir, ";
                     }
                 } else {
-                    $sortterm = $this->extraSortFields.', ';
+                    $sortterm = $this->extraSortFields . ', ';
                 }
             }
-            if ($list instanceof ArrayList) {
```
catharsisjelly, line 390: Since this is not backward compatible I thought it best to remove this `if` statement

src/GridFieldOrderableRows.php
```diff
@@ -384,29 +383,25 @@ public function getManipulatedData(GridField $grid, SS_List $list)
...
-                    $classname = $list->dataClass();
-                    if ($defaultSort = Config::inst()->get($classname, 'default_sort')) {
-                        if (is_array($defaultSort)) {
-                            $defaultSortArray = [];
-                            foreach ($defaultSort as $column => $direction) {
-                                $defaultSortArray[] = "\"$column\" $direction";
-                            }
-                            $defaultSort = implode(', ', $defaultSortArray);
+
+            $sortterm .= '"' . $this->getSortField() . '"';
```
catharsisjelly, line 390: Changing to v5 stuff the `$sortterm` var no longer requires the `getSortTable` call. That is unless I have misunderstood how this works (entirely possible). The tests were breaking without this change and now work if that helps

src/GridFieldOrderableRows.php
```diff
@@ -579,11 +574,12 @@ protected function executeReorder(GridField $grid, $sortedIDs)
                     $sortterm .= "$col $dir, ";
                 }
             } else {
-                $sortterm = $this->extraSortFields.', ';
+                $sortterm = $this->extraSortFields . ', ';
             }
         }
         $list = $grid->getList();
-        $sortterm .= '"'.$this->getSortTable($list).'"."'.$sortField.'"';
+        $sortterm .= '"' . $sortField . '"';
```
catharsisjelly, line 581: Changing to v5 stuff the `$sortterm` var no longer requires the `getSortTable` call. That is unless I have misunderstood how this works (entirely possible). The tests were breaking without this change and now work if that helps

lozcalver:
There is already a branch that’s SS5 compatible 😅 https://github.com/symbiote/silverstripe-gridfieldextensions/tree/4

catharsisjelly:
Awwww... nuts... didn'tt see that. Nuking this branch then

Issue #371, opened 2023-08-13
GridFieldEditableColumns does not support many_many through relationship's extra fields


rasstislav:
Hello,

I have this issue.

This module does not support many_many through relationship joined on a separate DataObject (https://docs.silverstripe.org/en/4/developer_guides/model/relations/#many-many-through-relationship-joined-on-a-separate-dataobject).

## PRs
- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/372 (didn't fix it apparently)

GuySartorelli:
Hi rasstislav,

Can you please provide some details?
What happens when you try, are there errors? If so what are the errors including the stack trace?
What version of the module and of silverstripe/framework are you seeing this problem on?
What specific component(s) are you using? And what were you doing at the time of the failure? This module comes with a lot of components and I'd be shocked if they _all_ don't work with a many_many through relationship.

rasstislav:
Hi @GuySartorelli,

Thank you for your response.

Module supports only standard [ManyManyList](https://github.com/silverstripe/silverstripe-framework/blob/4.13.18/src/ORM/ManyManyList.php), but not [ManyManyThroughList](https://github.com/silverstripe/silverstripe-framework/blob/4.13.18/src/ORM/ManyManyThroughList.php).

You can see e.g. here https://github.com/symbiote/silverstripe-gridfieldextensions/blob/3.6.3/src/GridFieldEditableColumns.php#L253.

GuySartorelli:
Hi @rasstislav 
Thank you for that. Please answer all of the questions I asked in my comment, including whether this is limited to just one or some components, or if it's actually _all_ components in this module which don't support that relation type.

rasstislav:
I am only using the GridFieldEditableColumns component and for [Ranking field](https://docs.silverstripe.org/en/4/developer_guides/model/relations/#many-many-through-relationship-joined-on-a-separate-dataobject) is displayed [ReadonlyField](https://github.com/symbiote/silverstripe-gridfieldextensions/blob/3.6.3/src/GridFieldEditableColumns.php#L280).

**Versions:**
silverstripe/framework: 4.13.13
symbiote/silverstripe-gridfieldextensions: 3.6.3

**Code:**

```
class Team extends DataObject
{
    private static $db = [
       'Name' => 'Varchar(35)',
    ];

    private static $many_many = [
        'Supporters' => [
            'through' => TeamSupporter::class,
            'from' => 'Team',
[...]

PR #372, opened 2023-08-14
FIX Allow editing extra fields from ManyManyThroughList.

GuySartorelli:
## Issue
- https://github.com/symbiote/silverstripe-gridfieldextensions/issues/371

### DIFF
--- a/src/GridFieldEditableColumns.php
+++ b/src/GridFieldEditableColumns.php
@@ -24,6 +24,7 @@ use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectInterface;
 use SilverStripe\ORM\ManyManyList;
+use SilverStripe\ORM\ManyManyThroughList;
 
 /**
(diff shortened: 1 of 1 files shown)


src/GridFieldEditableColumns.php
```diff
@@ -250,7 +251,7 @@ public function getFields(GridField $grid, DataObjectInterface $record)
                 }
             }
 
-            if (!$field && $list instanceof ManyManyList) {
+            if (!$field && ($list instanceof ManyManyList) || ($list instanceof ManyManyThroughList)) {
```
lozcalver:
```suggestion
            if (!$field && ($list instanceof ManyManyList || $list instanceof ManyManyThroughList)) {
```

I think this change currently runs the risk of passing this check even when `$field` has already been set:
```php
if (false && (true) || (false)) {
```

lozcalver (changes requested):

src/GridFieldEditableColumns.php
ChloeSartorelli: Good point. Done

Issue #383, opened 2024-04-04
Nested gridfield


maxime-rainville:
### Description

Our friends at [Creamarketing](https://www.creamarketing.com/) have developed a cool Nested GridField extension that allows users to manage a sub relation directly within a parent gridfield.

This is a pretty cool feature. Creamarketing has offered to contribute it back to this module.

![image](https://github.com/symbiote/silverstripe-gridfieldextensions/assets/1168676/1c6fe45d-2162-484f-b39f-e0b18e0204ed)

### User story
As a content author, I want a way to access nested relations in a GrdiField without having to drill down to individual records.

### Acceptance criteria
- Nested relations can be displayed in a secondary GridField within a primary GridField.
- Nested GridField can work with Has_many and many_many relations.
- The solution can work in both ModelAdmin and GridFields in the edit form of a record.
- Developers have an API for exposing specific relations in a nested Gridfield.
- All GridField components in framework and this module can work with Nested GridField.
  - The Orderable gridfiled component works within a single nested gridfield.
- All permission checks on DataObjects are retained inside nested GridFields.
- Code base is targeted to the namespace of the `symbiote/silverstripe-gridfieldextensions` module.
- Solution is functional with the Silverstripe CMS native GridField state management solution.
- Solution includes comprehensive PHPUnit tests.
- All user facing string are localised and provide a default English translation.
- The nested gridfield state when navigating between pages is retained.
- Nested Gridfield can be nested inside other nested GridFields to at least 10 levels.
- A maximum nested Gridfield recursion limit can be configured and has a sensible default value.
- DataObject with relations to themselves (e.g.Hierarchy) can be rendered recursively in a nested GridField (this could be used to display a tree in nested GridField) 
- The configuration for the nested GridField can be customised by a developer if needed.
- When a record doesn't have any nested relations, nested gridfield can be configured not to show.
  - (Default behaviour to run past the UX team)
- Nested GridField considers use case where a GridField is used to managed constructs that are not DataList/DataObjects. (Throwing meaningful exception is a valid solution for this AC)
- The GridField extensions developer docs are updated to explain how to use nested GridField.
- Developers have to op-in to using the feature.

### Excludes

Silverstripe will handle:
- Behat test.
- Risk analysis.
- Adding a nested GridField to frameworktest
- Design review (Maybe not ... need to clear this point with design team)

Other:
[...]

PR #384, opened 2024-04-15
NEW Nested gridfield

forsdahl:
## Description

Provides a `GridFieldComponent` for displaying a nested `GridField`, e.g. for `has_many` relations of the main record.

## Manual testing steps

Can be tested in a standard Silverstripe 5.1 installation in `SecurityAdmin` for editing groups, by adding the following extension to the project and adding it to `SecurityAdmin` via yaml config:

```php
<?php

namespace Creamarketing\NestedGridField;

use SilverStripe\Core\Extension;
use SilverStripe\Security\Group;
use Symbiote\GridFieldExtensions\GridFieldNestedForm;

class SecurityAdminExtension extends Extension
{
    public function updateGridFieldConfig($config)
    {
        if ($this->owner->getModelClass() === Group::class) {
            // shows nested groups
            $config->addComponent(GridFieldNestedForm::create());
            // Shows members in groups
            // $config->addComponent(GridFieldNestedForm::create()->setRelationName('Members'));
        }
    }
}
```

```yml
SilverStripe\Admin\SecurityAdmin:
  extensions:
    - Creamarketing\NestedGridField\SecurityAdminExtension
```

## Issues

- #383

## Pull request checklist

- [ ] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [ ] This change is covered with tests (or tests aren't necessary for this change)
- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [x] CI is green

### DIFF
--- a/src/GridFieldNestedForm.php
+++ b/src/GridFieldNestedForm.php
@@ -52,45 +52,78 @@ class GridFieldNestedForm extends GridFieldDetailForm implements
         $this->name = $name;
     }
     
+    /**
+     * Get the grid field that this component is attached to
+     * @return GridField
+     */
     public function getGridField()
     {
         return $this->gridField;
     }
 
+    /**
+     * Get the relation name to use for the nested grid fields
+     * @return string
+     */
     public function getRelationName()
     {
         return $this->relationName;
     }
     
+    /**
+     * Set the relation name to use for the nested grid fields
+     * @param string $relationName
+     */
     public function setRelationName($relationName)
     {
         $this->relationName = $relationName;
         return $this;
     }
 
+    /**
+     * Get whether the nested grid fields should be inline editable
+     * @return boolean
+     */
     public function getInlineEditable()
     {
         return $this->inlineEditable;
     }
 
+    /**
+     * Set whether the nested grid fields should be inline editable
+     * @param boolean $editable
+     */
     public function setInlineEditable($editable)
     {
         $this->inlineEditable = $editable;
         return $this;
     }
     
+    /**
+     * Set whether the nested grid fields should be expanded by default
+     * @param boolean $expandNested
+     */
     public function setExpandNested($expandNested)
     {
         $this->expandNested = $expandNested;
         return $this;
     }
 
+    /**
+     * Set whether the nested grid fields should be forced closed on load
+     * @param boolean $forceClosed
+     */
     public function setForceClosedNested($forceClosed)
     {
         $this->forceCloseNested = $forceClosed;
         return $this;
     }
 
+    /**
+     * Set a callback to check which items in this grid that should show the expand link
+     * for nested gridfields. The callback should return a boolean value.
+     * @param callable $checkFunction
+     */
     public function setCanExpandCheck($checkFunction)
     {
         $this->canExpandCheck = $checkFunction;
@@ -186,7 +219,7 @@ class GridFieldNestedForm extends GridFieldDetailForm implements
         if ($id) {
             // should be possible either on parent or child grid field, or nested grid field from parent
             $parent = $to ? $list->byID($to) : null;
-            if (!$parent && $to && $gridField->getForm()->getController() instanceof GridFieldNestedForm_ItemRequest && $gridField->getForm()->getController()->getRecord()->ID == $to) {
+            if (!$parent && $to && $gridField->getForm()->getController() instanceof GridFieldNestedFormItemRequest && $gridField->getForm()->getController()->getRecord()->ID == $to) {
                 $parent = $gridField->getForm()->getController()->getRecord();
             }
             $child = $list->byID($id);
@@ -248,7 +281,7 @@ class GridFieldNestedForm extends GridFieldDetailForm implements
         }
         $this->gridField = $gridField;
         $this->record = $record;
-        $itemRequest = new GridFieldNestedForm_ItemRequest($gridField, $this, $record, $gridField->getForm()->getController(), $this->name);
+        $itemRequest = GridFieldNestedFormItemRequest::create($gridField, $this, $record, $gridField->getForm()->getController(), $this->name);
         if ($request) {
             $pjaxFragment = $request->getHeader('X-Pjax');
             $targetPjaxFragment = str_replace('\\', '-', get_class($record)).'-'.$record->ID.'-'.$this->relationName;
--- a/src/GridFieldNestedForm.php
+++ b/src/GridFieldNestedForm.php
@@ -6,6 +6,7 @@ use Exception;
 use SilverStripe\Admin\ModelAdmin;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Core\Config\Configurable;
@@ -38,6 +39,9 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
 {
     use Configurable, GridFieldStateAware;
     
+    /**
+     * The key used in the post data to identify nested form data
+     */
     const POST_KEY = 'GridFieldNestedForm';
 
     private static $allowed_actions = [
@@ -54,15 +58,42 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
     /**
      * @var string
      */
-    protected $name;
-    protected $expandNested = false;
-    protected $forceCloseNested = false;
-    protected $gridField = null;
-    protected $record = null;
-    protected $relationName = 'Children';
-    protected $inlineEditable = false;
-    protected $canExpandCheck = null;
-    protected $maxNestingLevel = null;
+    private $name;
+
+    /**
+     * @var bool
+     */
+    private $expandNested = false;
+
+    /**
+     * @var bool
+     */
+    private $forceCloseNested = false;
+
+    /**
+     * @var GridField
+     */
+    private $gridField = null;
+
+    /**
+     * @var string
+     */
+    private $relationName = 'Children';
+
+    /**
+     * @var bool
+     */
+    private $inlineEditable = false;
+
+    /**
+     * @var callable|string
+     */
+    private $canExpandCheck = null;
+
+    /**
+     * @var int
+     */
+    private $maxNestingLevel = null;
     
     public function __construct($name = 'NestedForm')
     {
@@ -137,13 +168,14 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
     }
 
     /**
-     * Set a callback to check which items in this grid that should show the expand link
+     * Set a callback function to check which items in this grid that should show the expand link
      * for nested gridfields. The callback should return a boolean value.
-     * @param callable $checkFunction
+     * You can either pass a callable or a method name as a string.
+     * @param callable|string $callback
      */
-    public function setCanExpandCheck($checkFunction)
+    public function setCanExpandCheck($callback)
     {
-        $this->canExpandCheck = $checkFunction;
+        $this->canExpandCheck = $callback;
         return $this;
     }
 
@@ -212,7 +244,6 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
         $gridField->addExtraClass('has-nested');
         if ($record->ID && $record->exists()) {
             $this->gridField = $gridField;
-            $this->record = $record;
             $relationName = $this->getRelationName();
             if (!$record->hasMethod($relationName)) {
                 return '';
@@ -224,7 +255,7 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
                     return '';
                 } elseif (is_string($this->canExpandCheck)
                     && $record->hasMethod($this->canExpandCheck)
-                    && !$this->record->{$this->canExpandCheck}($record)
+                    && !$record->{$this->canExpandCheck}($record)
                 ) {
                     return '';
                 }
@@ -272,6 +303,11 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
         }
     }
 
+    /**
+     * Handle moving a record to a new parent
+     *
+     * @return string
+     */
     public function handleMoveToParent(GridField $gridField, $request)
     {
         $move = $request->postVar('move');
@@ -338,6 +374,14 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
         return $gridField->FieldHolder();
     }
     
+    /**
+     * Handle the request to show a nested item
+     *
+     * @param GridField $gridField
+     * @param HTTPRequest|null $request
+     * @param ViewableData|null $record
+     * @return HTTPResponse
+     */
     public function handleNestedItem(GridField $gridField, $request = null, $record = null)
     {
         if ($this->atMaxNestingLevel($gridField)) {
@@ -360,7 +404,6 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
             $gridField->getState(false)->setValue($gridStateStr);
         }
         $this->gridField = $gridField;
-        $this->record = $record;
         $itemRequest = GridFieldNestedFormItemRequest::create(
             $gridField,
             $this,
@@ -384,6 +427,13 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
         }
     }
 
+    /**
+     * Handle the request to toggle a nested item in the gridfield state
+     *
+     * @param GridField $gridField
+     * @param HTTPRequest|null $request
+     * @param ViewableData|null $record
+     */
     public function toggleNestedItem(GridField $gridField, $request = null, $record = null)
     {
         if (!$record) {
@@ -400,6 +450,12 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
         $state->$stateRelation = (int)$request->getVar('toggle');
     }
     
+    /**
+     * Get the link for the nested grid field
+     *
+     * @param string $action
+     * @return string
+     */
     public function Link($action = null)
     {
         $link = Director::absoluteURL(Controller::join_links($this->gridField->Link('nested'), $action));
@@ -407,6 +463,12 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
         return $manager->addStateToURL($this->gridField, $link);
     }
 
+    /**
+     * Get the link for the toggle action
+     *
+     * @param string $action
+     * @return string
+     */
     public function ToggleLink($action = null)
     {
         $link = Director::absoluteURL(Controller::join_links($this->gridField->Link('toggle'), $action, '?toggle='));
--- a/src/GridFieldNestedForm.php
+++ b/src/GridFieldNestedForm.php
@@ -104,27 +104,24 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
     
     /**
      * Get the grid field that this component is attached to
-     * @return GridField
      */
-    public function getGridField()
+    public function getGridField(): GridField
     {
         return $this->gridField;
     }
 
     /**
      * Get the relation name to use for the nested grid fields
-     * @return string
      */
-    public function getRelationName()
+    public function getRelationName(): string
     {
         return $this->relationName;
     }
     
     /**
      * Set the relation name to use for the nested grid fields
-     * @param string $relationName
      */
-    public function setRelationName($relationName)
+    public function setRelationName(string $relationName)
     {
         $this->relationName = $relationName;
         return $this;
@@ -132,18 +129,16 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
 
     /**
      * Get whether the nested grid fields should be inline editable
-     * @return boolean
      */
-    public function getInlineEditable()
+    public function getInlineEditable(): bool
     {
         return $this->inlineEditable;
     }
 
     /**
      * Set whether the nested grid fields should be inline editable
-     * @param boolean $editable
      */
-    public function setInlineEditable($editable)
+    public function setInlineEditable(bool $editable)
     {
         $this->inlineEditable = $editable;
         return $this;
@@ -151,9 +146,8 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
     
     /**
      * Set whether the nested grid fields should be expanded by default
-     * @param boolean $expandNested
      */
-    public function setExpandNested($expandNested)
+    public function setExpandNested(bool $expandNested)
     {
         $this->expandNested = $expandNested;
         return $this;
@@ -161,9 +155,8 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
 
     /**
      * Set whether the nested grid fields should be forced closed on load
-     * @param boolean $forceClosed
      */
-    public function setForceClosedNested($forceClosed)
+    public function setForceClosedNested(bool $forceClosed)
     {
         $this->forceCloseNested = $forceClosed;
         return $this;
@@ -173,9 +166,8 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
      * Set a callback function to check which items in this grid that should show the expand link
      * for nested gridfields. The callback should return a boolean value.
      * You can either pass a callable or a method name as a string.
-     * @param callable|string $callback
      */
-    public function setCanExpandCheck($callback)
+    public function setCanExpandCheck(callable|string $callback)
     {
         $this->canExpandCheck = $callback;
         return $this;
@@ -183,9 +175,8 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
 
     /**
      * Set the maximum nesting level allowed for nested grid fields
-     * @param int $level
      */
-    public function setMaxNestingLevel($level)
+    public function setMaxNestingLevel(int $level)
     {
         $this->maxNestingLevel = $level;
         return $this;
@@ -193,16 +184,14 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
 
     /**
      * Get the max nesting level allowed for this grid field.
-     * @return int
      */
-    public function getMaxNestingLevel()
+    public function getMaxNestingLevel(): int
     {
         return $this->maxNestingLevel ?: static::config()->get('default_max_nesting_level');
     }
 
     /**
      * Check if we are currently at the max nesting level allowed.
-     * @return bool
      */
     protected function atMaxNestingLevel(GridField $gridField): bool
     {
@@ -466,11 +455,8 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
     
     /**
      * Get the link for the nested grid field
-     *
-     * @param string $action
-     * @return string
      */
-    public function Link($action = null)
+    public function Link($action = null): string
     {
         $link = Director::absoluteURL(Controller::join_links($this->gridField->Link('nested'), $action));
         $manager = $this->getStateManager();
@@ -479,11 +465,8 @@ class GridFieldNestedForm extends AbstractGridFieldComponent implements
 
     /**
      * Get the link for the toggle action
-     *
-     * @param string $action
-     * @return string
      */
-    public function ToggleLink($action = null)
+    public function ToggleLink($action = null): string
     {
         $link = Director::absoluteURL(Controller::join_links($this->gridField->Link('toggle'), $action, '?toggle='));
         $manager = $this->getStateManager();
--- a/src/GridFieldNestedForm.php
+++ b/src/GridFieldNestedForm.php
@@ -9,6 +9,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
+use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Forms\GridField\AbstractGridFieldComponent;
 use SilverStripe\Forms\GridField\GridField;
@@ -21,6 +22,7 @@ use SilverStripe\Forms\GridField\GridFieldStateAware;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectInterface;
+use SilverStripe\ORM\Filterable;
 use SilverStripe\ORM\Hierarchy\Hierarchy;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\Versioned\Versioned;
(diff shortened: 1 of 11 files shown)


GuySartorelli:
Thanks for this!

I've added the pull request template to the PR description (did you remove that, or was it just not there when you started the PR?) and filled in some of it for you.
Can you please look at the rest of the checklist items and validate whether your PR meets those requirements? If it does, check the box. If it doesn't, update the PR until it does meet them.

I haven't looked at the code itself yet but I've added this to our review column.

GuySartorelli (changes requested):
I haven't done a comprehensive review of this, but here are some changes or questions I have so far:

- I think breadcrumbs in the GridField should include the parent record
    - In this case `Groups > Content Authors > Smith`
    image
- Search/filter functionality in nested gridfield doesn't seem to work
- Nested GridField state is not retained - e.g. if I apply a sort to the nested gridfield, click to edit a record, and then click back - the sort order is back to its default state.
- Seems to be no way to edit the nested GridField except through extensions
    - The acceptance criterion "the configuration for the nested GridField can be customised by a developer if needed" it a bit vague but IMO it is important to be able to customise it for *each* `GridField` independently, not just as a global extension
- There's an acceptance criterion in the issue about an upper limit for recursion, but I can't find where that's implemented - has that been implemented?
- "When a record doesn't have any nested relations, nested gridfield can be configured not to show."
    - Has this been implemented? If so please update the documentation to mention this.
    - Similarly if there are other options that haven't been documented, please document them so that I can test them and other developers can make use of them.
- The approach of subclassing `GridFieldDetailForm` seems a bit unusual, it's definitely not what I was expecting. Can you please briefly explain the reasons behind that decision?
- Please update the component list in the README.


I have been testing this by adding a nested gridfield to groups in `SecurityAdmin` with this code:

```php
<?php

namespace App\NestedExtensions;

use SilverStripe\Core\Extension;
use SilverStripe\Forms\GridField\GridFieldConfig;
use Symbiote\GridFieldExtensions\GridFieldNestedForm;

class NestedGroups extends Extension
{
    public function updateGridFieldConfig(GridFieldConfig $config)
    {
        $config->addComponent(GridFieldNestedForm::create()->setRelationName('Members'));
    }
}
```

```yml
SilverStripe\Admin\SecurityAdmin:
  extensions:
    - App\NestedExtensions\NestedGroups
```

forsdahl:
I now changed how Breadcrumbs work in nested GridFields. It now shows all the parent records, but the back link goes to the top-level instead of the parent record, since having to click back through all parent records to get back to the place you where is a bit confusing.

The reason `GridFieldNestedForm` extended `GridFieldDetailForm` are some legacy stuff, which isn't really true anymore in Silverstripe 5, so I now changed the base class to `AbstractGridFieldComponent` instead. `GridFieldNestedFormItemRequest` still extends `GridFieldDetailForm_ItemRequest` though, since those two actually share a lot of functionality.

The reason the search component doesn't work is probably due to the fact that the name of the nested GridFields contain "[" and "]" characters. The reason for this is so that `GridFieldEditableColumns` will work in a non-hacky way for nested GridFields. I'll see if I can come up with a solution that works for both components, or if I can find a fix for the search component working event with "["-characters in the GridField names.

I will implement some more UnitTests still, and look at the missing things for the acceptance criterias.

forsdahl:
I now changed the naming schema of the nested GridFields, to not include any "[" or "]" characters, since I got this schema to also work with `GridFieldEditableColumns`. This naming schema is safer, there might be other components also which do not like the brackets in the name.
`GridFieldOrderableRows` now also works with nested GridFields, both with immediate update, or without. For records with the `Hierarchy` extension, items can also be dragged between GridFields to move them. This can be tested with the following extension applied to `SecurityAdmin`:
```
<?php

namespace App\Extensions;

use SilverStripe\Core\Extension;
use SilverStripe\Security\Group;
use Symbiote\GridFieldExtensions\GridFieldNestedForm;
use Symbiote\GridFieldExtensions\GridFieldOrderableRows;

class SecurityAdminExtension extends Extension
{
    public function updateGridFieldConfig($config)
    {
        if ($this->owner->getModelClass() == Group::class) {
            $config->addComponent($nestedForm = new GridFieldNestedForm());
            $config->addComponent($orderableRows = new GridFieldOrderableRows());
            $orderableRows->setImmediateUpdate(true);
        }
    }
}
```

Also fixed some linting issues, so CI should now hopefully be green.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+    public function getMaxNestingLevel()
+    {
+        return $this->maxNestingLevel ?: $this->config()->max_nesting_level;
+    }
+
+    protected function getNestingLevel($gridField)
+    {
+        $level = 0;
+        $c = $gridField->getForm()->getController();
+        while ($c && $c instanceof GridFieldDetailForm_ItemRequest) {
```
ChloeSartorelli:
```suggestion
        while ($c && $c instanceof GridFieldNestedFormItemRequest) {
```
We only need to count nested gridfields - it doesn't matter if there are 2000 regular gridfield edit forms leading up to the root nested gridfield, those shouldn't be counted.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+            $c = $c->getController();
+            $level++;
+        }
+        return $level;
+    }
```
ChloeSartorelli: It is theoretically possible for an infinite loop of controllers referencing each other here - since we only need the nesting level to check if we're hitting the max, this method should be adjusted to return a boolean, which is whether we've hit the max nesting level or not. Something like this:

```php
private function atMaxNestingLevel(GridField $gridfield): bool
{
    $level = 0;
    $controller = $gridField->getForm()->getController();
    $maxLevel = $this->getMaxNestingLevel();
    while ($level < $maxLevel  && $controller && $controller instanceof GridFieldNestedFormItemRequest) {
        $controller = $controller->getController();
        $level++;
    }
    return $level >= $maxLevel;
}
```

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+     */
+    public function setMaxNestingLevel($level)
+    {
+        $this->maxNestingLevel = $level;
+        return $this;
+    }
+
+    public function getMaxNestingLevel()
+    {
+        return $this->maxNestingLevel ?: $this->config()->max_nesting_level;
```
ChloeSartorelli:
```suggestion
        return $this->maxNestingLevel ?: static::config()->get('max_nesting_level');
```
Config is a static method so should be accessed statically.
The get method should be explicitly used instead of relying on the magic `__get()`, to make it clearer what's happening.

If there's any other usage of config in this PR, please also update those.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+{
+    use Configurable, GridFieldStateAware;
+    
+    const POST_KEY = 'GridFieldNestedForm';
+
+    private static $allowed_actions = [
+        'handleNestedItem'
+    ];
+
+    private static $max_nesting_level = 10;
```
ChloeSartorelli:
```suggestion
    private static $default_max_nesting_level = 10;
```
This value is used as a default, so we should be explicit about that.

Please also add a PHPDoc for this since it will display in the API documentation. Same with any config you've added in this PR that isn't inherited.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+     */
+    protected $name;
+    protected $expandNested = false;
+    protected $forceCloseNested = false;
+    protected $gridField = null;
+    protected $record = null;
+    protected $relationName = 'Children';
+    protected $inlineEditable = false;
+    protected $canExpandCheck = null;
+    protected $maxNestingLevel = null;
```
ChloeSartorelli: Please add strict typehints for these (including the config properties), and for all methods that aren't being inherited from superclasses or interfaces. For all new API, we always add strict typehints as best practice.
Please see https://docs.silverstripe.org/en/5/contributing/coding_conventions/php_coding_conventions/

ChloeSartorelli: For all of these which have a `getter` and `setter` method, please make the property `private`. There's no reason for subclasses to be accessing the properties directly.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+        $this->name = $name;
+    }
+    
+    /**
+     * Get the grid field that this component is attached to
+     * @return GridField
+     */
+    public function getGridField()
+    {
+        return $this->gridField;
```
ChloeSartorelli:
```suggestion
    public function getGridField(): GridField
    {
        return $this->gridField;
```
Please add strict typehints for arguments and return types for all methods in this PR which aren't inherited from superclasses or interfaces.

Note that PHPDocs are only required for parameters and return types if there is additional information, such as a description. If it's just the type, we don't need the PHPDoc when there's a real typehint.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+            return '';
+        }
+        $gridField->addExtraClass('has-nested');
+        if ($record->ID && $record->exists()) {
+            $this->gridField = $gridField;
+            $this->record = $record;
+            $relationName = $this->getRelationName();
+            if (!$record->hasMethod($relationName)) {
+                return '';
+            }
```
ChloeSartorelli: This strikes me as a time we should be throwing an exception. If the developer has said "this relation exists", we shouldn't fail silently if we discover it doesn't actually exist.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+     */
+    public function setForceClosedNested($forceClosed)
+    {
+        $this->forceCloseNested = $forceClosed;
+        return $this;
+    }
+
+    /**
+     * Set a callback to check which items in this grid that should show the expand link
+     * for nested gridfields. The callback should return a boolean value.
```
ChloeSartorelli: The usage of `canExpandCheck` makes it clear that the name of a method on the parent record can also be passed in. Please update the PHPDoc to make it clear this is possible.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+     * @param callable $checkFunction
+     */
+    public function setCanExpandCheck($checkFunction)
```
ChloeSartorelli:
```suggestion
     */
    public function setCanExpandCheck(callable|string $checkFunction): static
```

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+    {
+        $this->canExpandCheck = $checkFunction;
```
ChloeSartorelli:
```suggestion
    public function setCanExpandCallback($callback)
    {
        $this->canExpandCallback = $callback;
```
Just making it clearer by the naming what this actually is - the property holds a callback (or the name of a method).

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+                return '';
+            }
+            if ($this->canExpandCheck) {
+                if (is_callable($this->canExpandCheck)
+                    && !call_user_func($this->canExpandCheck, $record)
+                ) {
+                    return '';
+                } elseif (is_string($this->canExpandCheck)
+                    && $record->hasMethod($this->canExpandCheck)
+                    && !$this->record->{$this->canExpandCheck}($record)
```
ChloeSartorelli: - We don't need the `is_string()` check since the setter method can only accept a string or a callable.
- We should throw an exception if the record doesn't have that method, since that scenario means the developer has misconfigured this component. Not throwing an exception is just silently failing.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+            'POST movetoparent' => 'handleMoveToParent'
+        ];
+    }
+
+    /**
+     * @param GridField $field
+     */
+    public function getHTMLFragments($field)
+    {
+        if (DataObject::has_extension($field->getModelClass(), Hierarchy::class)) {
```
ChloeSartorelli:
```suggestion
        if (is_a($field->getModelClass(), DataObject::class, true) && DataObject::has_extension($field->getModelClass(), Hierarchy::class)) {
```
Don't assume the modelclass is a `DataObject` subclass.
Please check the PR for other scenarios where that assumption has been made.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+            'nested/$RecordID/$NestedAction' => 'handleNestedItem',
+            'toggle/$RecordID' => 'toggleNestedItem',
+            'POST movetoparent' => 'handleMoveToParent'
+        ];
+    }
+
+    /**
+     * @param GridField $field
+     */
+    public function getHTMLFragments($field)
```
ChloeSartorelli: The name of this method implies it should be returning something. Have you missed a return statement? Or are we (ab)using this method for something it wasn't intended for?

If the latter, at a minimum please include a code statement explaining what's going on, but preferably find a more suitable place to put the logic.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+    {
+        if (DataObject::has_extension($field->getModelClass(), Hierarchy::class)) {
+            $field->setAttribute('data-url-movetoparent', $field->Link('movetoparent'));
+        }
+    }
+
+    public function handleMoveToParent(GridField $gridField, $request)
```
ChloeSartorelli: Please add a PHPDoc comment to this and to all other public methods in this PR which aren't inherited from superclasses or interfaces.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,435 @@
...
+
+                        /** @var GridFieldOrderableRows */
+                        $orderableRows = $gridField->getConfig()->getComponentByType(GridFieldOrderableRows::class);
+                        if ($orderableRows) {
+                            $orderableRows->setImmediateUpdate(true);
+                            try {
+                                $orderableRows->handleReorder($gridField, $request);
+                            } catch (Exception $e) {
+                            }
+                        }
```
ChloeSartorelli: This level of nested control statements is a bit of a code smell - it's quite hard to follow the flow of logic at a glance. Please try to restructure this so that we're exiting early where possible. It may be appropriate to refactor some of this logic into private methods instead of having so many nested control statements.

src/GridFieldNestedFormItemRequest.php
```diff
@@ -0,0 +1,173 @@
...
+use SilverStripe\Forms\GridField\GridFieldSortableHeader;
+use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\HasManyList;
+use SilverStripe\ORM\Hierarchy\Hierarchy;
+use SilverStripe\View\ArrayData;
+use Symbiote\GridFieldExtensions\GridFieldAddNewInlineButton;
+use Symbiote\GridFieldExtensions\GridFieldEditableColumns;
+use Symbiote\GridFieldExtensions\GridFieldOrderableRows;
+
+class GridFieldNestedFormItemRequest extends GridFieldDetailForm_ItemRequest
```
ChloeSartorelli, line 30: Please add a PHPDoc comment to this and to all public methods in this class which aren't inherited from the superclass.

src/GridFieldNestedFormItemRequest.php
```diff
@@ -0,0 +1,173 @@
...
+{
+
```
ChloeSartorelli:
```suggestion
```

src/GridFieldNestedFormItemRequest.php
```diff
@@ -0,0 +1,173 @@
...
+class GridFieldNestedFormItemRequest extends GridFieldDetailForm_ItemRequest
+{
+    
+    public function Link($action = null)
+    {
+        return Controller::join_links($this->component->Link($this->record->ID), $action);
+    }
+    
+    public function ItemEditForm()
```
ChloeSartorelli, line 37: Please refactor this a bit so that the nested gridfield's `GridFieldConfig` is passed in from the `GridFieldNestedForm` component.
Then, please add a getter and setter method on `GridFieldNestedForm` so that developers have direct access to the config which will be used in the nested gridfield.

i.e. it should be possible to do this (feel free to change the method names):
```php
$nestedGridfieldComponent = new GridFieldNestedForm();
$parentGridFieldConfig->addComponent($nestedGridfieldComponent);

// alter the nested gridfield config
$nestedConfig = $nestedGridfieldComponent->getNestedConfig();
$nestedConfig->// add or remove components to suit business needs

// completely replace the nested gridfield config
$nestedGridfieldComponent->setNestedConfig($myNestedConfig);
```

src/GridFieldNestedFormItemRequest.php
```diff
@@ -0,0 +1,173 @@
...
+        );
+        if (!trim($title)) {
+            $gridField->addExtraClass('empty-title');
+        }
+        $gridField->setModelClass($relationClass);
+        $gridField->setAttribute('data-class', str_replace('\\', '-', $relationClass));
+        $gridField->addExtraClass('nested');
+        $form = new Form($this, 'ItemEditForm', $fields, new FieldList());
+        $form->setStrictFormMethodCheck(false);
+        $form->disableSecurityToken();
```
ChloeSartorelli: Why are we disabling the security token and strict method check?

templates/Symbiote/GridFieldExtensions/GridFieldNestedForm.ss
```diff
@@ -0,0 +1,6 @@
+<a class="btn btn-secondary btn--no-text btn--icon-large <% if $Toggle == 'open' %>font-icon-down-dir<% else %>font-icon-right-dir<% end_if %> cms-panel-link list-children-link" data-pjax-target="$PjaxFragment" href="$Link" data-toggle="$ToggleLink"></a>
```
ChloeSartorelli: Do I understand correctly that this "link" is used to collapse/expand the nested gridfield? If so please use the appropriate `aria` attributes for accessibility.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded should set you on the right track.

ChloeSartorelli: Also please use a `button` element instead of a link if possible. Buttons are more semantically correct here (and don't have the unexpected behaviour when middle clicking)

GuySartorelli (changes requested):
I haven't done a full thorough code review but I have taken a look and highlighted some areas that don't meet our coding standards or which seemed like they could be improved.

I've also done another in-browser test and these are the problems I encountered:

- There's some unexpected state conflict between the nested gridfield and the main gridfield. Steps to reproduce:
    1. Have at least two items in the main gridfield, both with nested gridfields
    1. expand both nested gridfields
    1. Filter the main gridfield (e.g. search for "author") so that there's only one item in the main gridfield.
    1. Filter the nested gridfield.
    1. Click on a record in the nested gridfield, then click the back arrow in the breadcrumbs
    1. Remove the filter on the *main* gridfield
    1. Click on a record in the nested gridfield, then click the back arrow in the breadcrumbs
    1. Notice that the filter (which you cleared just before) is back again.
- Calling `setMaxNestingLevel(3)` on the nested gridfield component doesn't seem to work - though changing the configuration does.
    - My guess is we need to pass the nesting level down recursively to all child nested gridfield components. We should also disallow developers setting the value in nested components directly, so we don't get a parent with a max nesting level of 3 and a child with a max nesting level of 10 (as a random example).
    - I would have expected the nested gridfield settings to be inherited automatically, since this was tested with a `Hierarchy` object (using the example in the PR description). It looks like the reason this didn't happen is because the nested gridfield in this case is set up in `GridFieldNestedFormItemRequest` as a fresh instance of `GridFieldNestedForm()` which doesn't look at the settings defined in `getCMSFields()` at all.
- When trying to use sort to move items between nested gridfields, I moved a child group to be a sibling to what was its parent group. This resulted in a popup that said "An error occured while fetching data from the server Please try again later." - and an error in the network tab of the developer console which said `ERROR [UNKNOWN TYPE, ERRNO 400]`. The full error with stack trace was:
    ```text
    ERROR [UNKNOWN TYPE, ERRNO 400]: 
    IN POST /admin/security/groups/EditForm/field/groups/nested/1/ItemEditForm/field/groups-GridFieldNestedForm-1/nested/3/ItemEditForm/field/groups-GridFieldNestedForm-1-GridFieldNestedForm-3/nested/5/ItemEditForm/field/groups-GridFieldNestedForm-1-GridFieldNestedForm-3-GridFieldNestedForm-5/reorder?gridState-groups-0=%7B%22GridFieldOrderableRows%22%3A%7B%22enabled%22%3Atrue%7D%2C%22GridFieldNestedForm%22%3A%7B%22SilverStripe-Security-Group-2-Children%22%3Anull%2C%22SilverStripe-Security-Group-1-Children%22%3A1%7D%7D&gridState-groups-GridFieldNestedForm-1-1=%7B%22GridFieldOrderableRows%22%3A%7B%22enabled%22%3Atrue%7D%2C%22GridFieldNestedForm%22%3A%7B%22SilverStripe-Security-Group-3-Children%22%3A1%7D%7D&gridState-groups-GridFieldNestedForm-1-GridFieldNestedForm-3-2=%7B%22GridFieldOrderableRows%22%3A%7B%22enabled%22%3Atrue%7D%2C%22GridFieldNestedForm%22%3A%7B%22SilverStripe-Security-Group-4-Children%22%3Anull%2C%22SilverStripe-Security-Group-5-Children%22%3A1%7D%7D
    Line  in 
    
    Trace
    =====
    SilverStripe\Logging\DetailedErrorFormatter->output(400, , , , )
    DetailedErrorFormatter.php:55
    
    SilverStripe\Logging\DetailedErrorFormatter->format(Array)
    HTTPResponse.php:417
    
    SilverStripe\Control\HTTPResponse->outputBody()
    HTTPResponse.php:346
    
    SilverStripe\Control\HTTPResponse->output()
    index.php:25
    ```
    - It's worth noting that the group *did* move to where it was meant to move to, despite the error.
    
Additionally, please do these actions I previously requested in https://github.com/symbiote/silverstripe-gridfieldextensions/pull/384#pullrequestreview-2007670453
- Please add additional documentation so I know what needs to be manually tested and so that developers know what functionality this thing has.
- Please update the component list in the readme

src/GridFieldNestedForm.php
forsdahl: Actually, the way the code works now it is needed. The first if statement will return false if it is a callable but the result from this callable is true (i.e. we can expand nested items on this record). So either it should be as it is, or it needs to be refactored a bit.

src/GridFieldNestedForm.php
forsdahl: Also added some checks to not assume the underlying list is Filterable, i.e. don't assume it has `filter` or `byID` functions.

src/GridFieldNestedForm.php
forsdahl: This function is used because of the move between gridfields functionality. That functionality currently needs the move-to-parent url as a data-attribute on the grid field. This attribute needs to be added before anything is rendered for the gridfield, and it is also not dependent on individual records, so `getColumnContent` is also not optimal. The optimal place to do it seems to be `getURLHandlers`, but that is called too late, i.e. the added attribute will not be rendered. Added a comment about this in the code.

src/GridFieldNestedForm.php
forsdahl: You are completely right, this code is not good. The logic is a bit complicated here, since there are quite a few possible cases for the moving functionality, but I now at least refactored it to use early return instead of this deep control statements structure.

src/GridFieldNestedFormItemRequest.php
forsdahl: This is some legacy stuff that doesn't seem to be needed anymore (probably dates back to before Silverstripe 4). I now removed these, all functionality seem to work without these, and better to leave them on than disabling them.

src/GridFieldNestedFormItemRequest.php
forsdahl: One reason the gridfield config is created where it is now is that in the default config, we sometimes want to check permissions on the specific record to see what components should be added. For example the default config is based on `GridFieldConfig_RecordEditor`, but if the user can't edit a given record, then that record's nested gridfield will not include the add button. This won't be possible in the same way if we move the logic to `GridFieldNestedForm`, since then the config is the same for all records. For now we have found it more usable to us to configure the nested gridfield config directly in the DataObject classes, or via extensions to said classes.
I can refactor it this way, but then we need to deside if we do any of these permission checks at all, since they might then be confusing for the developer.

templates/Symbiote/GridFieldExtensions/GridFieldNestedForm.ss
forsdahl: Good points, much better with a button with aria-expanded instead of an anchor link.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,517 @@
...
+
+    /**
+     * @var callable|string
+     */
+    private $canExpandCheck = null;
+
+    /**
+     * @var int
+     */
+    private $maxNestingLevel = null;
```
ChloeSartorelli:
```suggestion
     */
    private static bool $default_max_nesting_level = 10;
    
    private string $name;

    private bool $expandNested = false;

    private bool $forceCloseNested = false;

    private GridField $gridField = null;

    private string $relationName = 'Children';

    private bool $inlineEditable = false;

    private callable|string $canExpandCheck = null;

    private int $maxNestingLevel = null;
```

ChloeSartorelli: As per https://github.com/symbiote/silverstripe-gridfieldextensions/pull/384#discussion_r1573963964 these should be strict typehints - i.e. PHP should thrown errors at us if we set values that aren't those types.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,517 @@
...
+     */
+    public function getInlineEditable(): bool
+    {
+        return $this->inlineEditable;
+    }
+
+    /**
+     * Set whether the nested grid fields should be inline editable
+     */
+    public function setInlineEditable(bool $editable)
```
ChloeSartorelli:
```suggestion
    public function setInlineEditable(bool $editable): static
```
Please do this for all other `set*` methods which return `$this`

src/GridFieldNestedForm.php
ChloeSartorelli: You marked this as resolved but didn't make any change, nor did you reply. Is there a reason to not throw an exception here? If not please make this change.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,517 @@
...
+    }
+
+    /**
+     * Set a callback function to check which items in this grid that should show the expand link
+     * for nested gridfields. The callback should return a boolean value.
+     * You can either pass a callable or a method name as a string.
+     */
+    public function setCanExpandCheck(callable|string $callback)
+    {
+        $this->canExpandCheck = $callback;
```
ChloeSartorelli:
```suggestion
        $this->canExpandCallback = $callback;
```
As per https://github.com/symbiote/silverstripe-gridfieldextensions/pull/384#discussion_r1573966303

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,517 @@
...
+        }
+        return [];
+    }
+
+    /**
+     * Handle moving a record to a new parent
+     *
+     * @return string
+     */
+    public function handleMoveToParent(GridField $gridField, $request)
```
ChloeSartorelli:
```suggestion
     */
    public function handleMoveToParent(GridField $gridField, $request): string
```
For all methods which are not inherited from a parent class or interface, please use strict typehints instead of phpdocs, e.g. handleNestedItem. I won't list them all.

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,517 @@
...
+            'nested/$RecordID/$NestedAction' => 'handleNestedItem',
+            'toggle/$RecordID' => 'toggleNestedItem',
+            'POST movetoparent' => 'handleMoveToParent'
+        ];
+    }
+
+    /**
+     * @param GridField $gridField
+     * @return array
+     */
```
ChloeSartorelli: This is defined on the interface, so it isn't needed here.
```suggestion
```

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,517 @@
...
+    public function getHTMLFragments($gridField)
+    {
+        /**
```
ChloeSartorelli:
```suggestion
        /*
```
The double asterisk tells IDEs that this comment may include typehints etc - but this is just a multi-line comment so it shouldn't have that.
I'm nitpicking here - I won't hold back merging if this change isn't made.

src/GridFieldNestedForm.php
ChloeSartorelli: Yup good point, I must not have had my coffee when I looked at this bit :p

src/GridFieldNestedForm.php
```diff
@@ -0,0 +1,517 @@
...
+     * @return string
+     */
+    public function handleMoveToParent(GridField $gridField, $request)
+    {
+        $move = $request->postVar('move');
+        /** @var DataList */
+        $list = $gridField->getList();
+        $id = isset($move['id']) ? (int) $move['id'] : null;
+        if (!$id) {
+            throw new HTTPResponse_Exception('Missing ID', 400);
```
ChloeSartorelli:
```suggestion
            throw new HTTPResponse_Exception('Missing ID', 404);
```
404 if the item is not found

src/GridFieldNestedFormItemRequest.php
ChloeSartorelli: Ahh I've just noticed the `getNestedConfig()` check there which solves some of this. That should definitely be added to the documentation (which still needs to be added to as per https://github.com/symbiote/silverstripe-gridfieldextensions/pull/384#pullrequestreview-2013562592 "Please add additional documentation so I know what needs to be manually tested and so that developers know what functionality this thing has.").

That's probably okay for now, given there is at least a per-model level of control.

GuySartorelli (changes requested):
Thank you for your patience and diligent effort on this so far, we're very close now.

Please avoid marking comments as "resolved" - I use them to know what changes I requested last time (and have verified changes were made) vs which changes have already been dealt with in previous reviews.

There's also a minor PHP linting issue to resolve.

GuySartorelli:
In addition to the above, the following from https://github.com/symbiote/silverstripe-gridfieldextensions/pull/384#pullrequestreview-2013562592 haven't been resolved:

- There's some unexpected state conflict between the nested gridfield and the main gridfield (steps to reproduce in the original comment)
- Please add additional documentation so I know what needs to be manually tested and so that developers know what functionality this thing has.
- Please update the component list in the readme

GuySartorelli:
I've discovered that if you try to add two `GridFieldNestedForm` components, it doesn't work as expected (e.g. uncomment the second component in the PR description).
Probably we should just throw an exception if trying to add a `GridFieldNestedForm` when there's already one there. We can look at having multiple nested sibling gridfields working nicely together as a follow-up, if there's appetite for that.

forsdahl:
I looked in to the state handling issues you described, and it does seem to me to be the default state handling logic which isn't really up to handling this. The `GridFieldNestedForm` component doesn't do any custom logic for the state handling, neither in php nor in javascript, it relies on the gridfield state manager to add the correct state parameters for all links.
I can reproduce the scenario which you describe, and I can see that when having both a base grid field and a nested grid field with the `GridFieldFilterHeader` component, the state-part of the url is not updated properly when filtering.
I think the solution here is to update the general gridfield state handling to support this, with the `GridFieldStateManager` implementation we have which stores individual gridfield states in session, and only adds a state key to the url.
In the mean time, I think we could at least disable the `GridFieldFilterHeader` in the default nested gridfield config. In all the cases we have used nested gridfields, we have yet to see a real-world use case where we really feel the filtering in the nested gridfield to be really necessary and I suspect the state handling works ok if only the base gridfield has the `GridFieldFilterHeader`-component.

GuySartorelli:
There are still outstanding items I've requested above so I'll assume you're still working on this.
I've let CI run so you can see if that's green in the meantime.

src/GridFieldNestedForm.php
forsdahl: I changed these as requested, except the `$canExpandCallback` property. Seems like php doesn't like `callable` as a strict typehint on class properties, so I left that as a docblock typehint.

GuySartorelli:
To find it easier to find the requested changes that haven't been made yet:
- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/384#issuecomment-2089247085 regarding documentation and the readme
- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/384#issuecomment-2089247995

forsdahl:
Documentation and readme now updated.
GridFieldNestedForm now also throws an exception if it detects more than one instance of itself on the same GridField.

GuySartorelli (approved):
LGTM, nothing stands out as being obvious wrong in the code and it works great locally.
Thanks for your efforts on this, it's a really fantastic contribution!

This feature will be included in the October minor release.

Issue #403, opened 2024-06-06
Nested gridifeld forces open the left sidebar


maxime-rainville:
When navigating nested GridFlieds, the left side bar is forced open, even if you set it to collapse.

If you toggle the "fix" option it stays closes.

https://youtu.be/-vPjylr98fQ

## Step to reproduce
- Navigate to a a nested gridfield model admin with at least one entry
- Collapse your side bar without choosing the "fixed" option.
- Expand an entry the gridfield.
 
Expected: The entry is expanded and the sidebar stays collapse.
Actual: The entry is loaded, but the collapse sidebar is expanded.

## Acceptance criteria
- The sidebar stays collapse when drilling down a nested gridfield whether "fixed" is collapse or not.

## PRs
- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/406

GuySartorelli:
PR merged

PR #406, opened 2024-06-21
FIX Prevent popping open cms-menu when using nested gridfield

emteknetnz:
Issue https://github.com/symbiote/silverstripe-gridfieldextensions/issues/403

In LeftAndMain there's a handler for [$(document).ajaxComplete()](https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/legacy/LeftAndMain.js#L183) causing a redraw.  This is resolved by using fetch() instead of jquerys ajax function.

Requires https://github.com/silverstripe/silverstripe-frameworktest/issues/191 to be merged first

### DIFF
--- a/javascript/GridFieldExtensions.js
+++ b/javascript/GridFieldExtensions.js
@@ -550,34 +550,30 @@
 								data[parts[0]] = parts[1];
 							}
 						}
-						$.ajax({
-							type: 'POST',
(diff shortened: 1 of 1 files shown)


GuySartorelli (approved):
LGTM, works well locally.

PR #415, opened 2024-08-23
MNT Fix some unit tests

GuySartorelli:
Fixes some unit tests. Comments will explain why the changes were needed.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11322

### DIFF
--- a/tests/GridFieldEditableColumnsTest.php
+++ b/tests/GridFieldEditableColumnsTest.php
@@ -5,8 +5,6 @@ namespace Symbiote\GridFieldExtensions\Tests;
 use Symbiote\GridFieldExtensions\Tests\Stub\TestController;
 use Symbiote\GridFieldExtensions\Tests\Stub\StubUnorderable;
 use Symbiote\GridFieldExtensions\GridFieldEditableColumns;
-use SilverStripe\ORM\FieldType\DBHTMLText;
-use SilverStripe\Forms\TextField;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\Form;
 use SilverStripe\Forms\FieldList;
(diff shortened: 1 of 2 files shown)


tests/OrderableRowsThroughVersionedTest.php
```diff
@@ -33,7 +33,7 @@ class OrderableRowsThroughVersionedTest extends SapphireTest
     protected function setUp(): void
     {
         parent::setUp();
-        $this->orignalReadingMode = Versioned::get_reading_mode();
+        $this->originalReadingMode = Versioned::get_reading_mode();
```
ChloeSartorelli, line 36: See line 31 above - this was a typo which resulted in a PHP deprecation warning in CI. Super minor, doesn't really matter until they actually remove the ability to have dynamic properties and it's only a unit test so I don't think it's worth backporting.

tests/GridFieldEditableColumnsTest.php
```diff
@@ -5,8 +5,6 @@
 use Symbiote\GridFieldExtensions\Tests\Stub\TestController;
 use Symbiote\GridFieldExtensions\Tests\Stub\StubUnorderable;
 use Symbiote\GridFieldExtensions\GridFieldEditableColumns;
-use SilverStripe\ORM\FieldType\DBHTMLText;
-use SilverStripe\Forms\TextField;
```
ChloeSartorelli, line 9: Neither of these are used

tests/GridFieldEditableColumnsTest.php
```diff
@@ -47,10 +45,9 @@ public function testProvidesEditableFieldsInColumns()
         $record->setCanEdit(true);
         $column = $component->getColumnContent($grid, $record, 'Title');
 
-        $this->assertInstanceOf(DBHTMLText::class, $column);
         $this->assertMatchesRegularExpression(
             '/<input type="text" name="TestGridField\[GridFieldEditableColumns\]\[100\]\[Title\]" value="foo"[^>]*>/',
-            $column->getValue()
+            $column
```
ChloeSartorelli, line 50: The value for the column comes from `forTemplate()` which is now typed as `string`.

emteknetnz (approved):

PR #418, opened 2024-09-21
FIX #355 Sorting in paginated lists

### LINKED ISSUE #355: BUG: GridFieldOrderableRows interaction with GridFieldPaginator
# BUG: GridFieldOrderableRows interaction with GridFieldPaginator

Impacted version: `3.5.0`

Re-ordering items across pages produces unexpected results.

## Steps to reproduce

* taxonomy group <- many many through (with `Sort` on through object) -> taxonomy term (also has a `Sort` column)
* generate a taxonomy group with 200 taxonomy terms
* navigate to a page (for example page `4`)
* drag an items to be reordered (for example item number `50`) and drop it on the previous page icon in the paginator

### Expected result

Item number `50` will be moved to the previous page while leaving all the items in the same place

### Actual result

The list gets completely shifted with first 44 items leaving their place on the first page.

#### Screenshots

Screen Shot 2023-01-23 at 8 31 59 AM

Screen Shot 2023-01-23 at 8 32 16 AM

Screen Shot 2023-01-23 at 8 33 25 AM

### Notes

I haven't done too much investigation into this but it's possible that the random sort order is coming from the `Sort` column on the taxonomy terms. Maybe the re-order action gets confused on which DB column needs to be used when using the paginator reorder.

Normal reorder action within a single page works fine though.

## PRs
- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/418

johannesx75:
## Description
If a many_many relation with the sort in an many_many_extraField has an item dragged to another page the resulting sort is wrong. This was caused by the handleMoveToPage function not respecting the original sort of the list.

## Issues
- #355 

## Pull request checklist

- [x] CI is green

### DIFF
--- a/src/GridFieldOrderableRows.php
+++ b/src/GridFieldOrderableRows.php
@@ -500,7 +500,7 @@ class GridFieldOrderableRows extends RequestHandler implements
         $move  = $request->postVar('move');
         $field = $this->getSortField();
 
-        $list  = $grid->getList();
+        $list  = $grid->getList()->sort($field);
         $manip = $grid->getManipulatedList();
 
         $existing = $manip->map('ID', $field)->toArray();


GuySartorelli:
Thank you for submitting this change.
Can you please retarget this to the `4.0` branch? Otherwise it'll need to wait until the April 2025 minor release.

You may need to [reset your commits](https://gist.github.com/GuySartorelli/74627843aeb75e12b777b1f44cb22d2f) after retargetting the PR.

johannesx75:
Ok, I've changed it to the 4.0 branch.

GuySartorelli:
Thanks for doing that.
I've just tested this PR and it worked fine. But I then tried to reproduce the original bug, and I can't. The reproduction steps there aren't very detailed about how to set up the relations or how to create the records.

Here's my class for the "taxonomy group" and the through class that the reproduction steps say I need:

```php

class TaxonomyGroup extends DataObject
{
    private static $many_many = [
        'Terms' => [
            'through' => TermThrough::class,
            'from' => 'Group',
            'to' => 'Term',
        ],
    ];
}
```


```php
use SilverStripe\Taxonomy\TaxonomyTerm;

class TermThrough extends DataObject
{
    private static $db = [
        'Sort' => 'Int',
    ];

    private static $has_one = [
        'Group' => TaxonomyGroup::class,
        'Term' => TaxonomyTerm::class,
    ];
}
```

And here's how I generated the records:

```php
use SilverStripe\Taxonomy\TaxonomyTerm;

if (TaxonomyGroup::get()->count() === 0) {
    $group = TaxonomyGroup::create();
    $group->write();

    for ($i = 0; $i < 200; $i++) {
        $term = TaxonomyTerm::create(['Name' => $i, 'Sort' => $i]);
        $term->write();
        $group->Terms()->add($term);
    }
}
```

With that setup I can't actually reproduce the original bug (with or without the PR).
Can you or @mfendeksilverstripe please let me know what I need to change to reproduce the original bug?

@mfendeksilverstripe can you please test this PR and check it resolves the problem for you in CMS 5.2+?

mfendeksilverstripe:
Hey @GuySartorelli I did test this change but it doesn't seem to be fixing the issue. I suspect that we might have two scenarios here to test: many_many_extra_field with sort and many_many_through with sort. I tested the latter.

This is my local test setup:

* parent model < many many through with sort > child model
* add GridField with the many many relation from parent model to child model
* make sure you have plenty of child models so you need pagination with multiple pages
* make sure that you have orderable rows component available in the GridField
* then scroll to the bottom of the page 1 and click the last item orderable component widget and drag it to the next page icon on the paginator (it should highlight itself as you hover over it), finally drop the item which should trigger reorder action

For me it reorders different items compared to what I selected. For example, I attempted to move item 15 to the next page and instead it moved item 3 to the last page.

If your still having trouble with the setup please let me know and I can assist you further as we can use one of my projects as a testing ground.

GuySartorelli:
@johannesx75 Sounds like there's still some work to don on this PR, please amend the PR as per the comment above.

mikeyc7m:
I have also hit this problem, on a hasMany list. Each time I send an item from page 2 back to page 1 the whole sort order gets scrambled. The fix in this PR does not help.
When the sort order is applied, all the items on page 2 get their numbering reset. So the sequence of the "Sort" field goes:
Item 16 becomes 0
Item 17 becomes 1, along with the previous Item 1.
Item 18 becomes 2, along with the previous item 2.

Issue #436, opened 2025-04-16
Using GridFieldConfigurablePaginator in a GridField in SilverStripe 5.4 removes all entries in GridField


Oliver996:
### Module version(s) affected

4.1.0

### Description

I have created my own GriedFieldConfig that we already use for a long time (since SilverStripe 3). This config uses GridFieldConfigurablePaginator, but now in SilverStripe 5.4 when I go to my GridField that uses this field, I can't see entries in the GridField list. It only says "no Items Found". But when I change GridFieldConfigurablePaginator to GridFieldPaginator, the list is displayed normally. I also tested it with GridFieldConfig_RecordEditor the same result: GridFieldConfigurablePaginator no items and GridFieldPaginator items in the list. Sadly, I did not really find a solution for this problem. Can you help me?

Also tested on a clean installation with only this modul.

### How to reproduce

```php
$config = GridFieldConfig_RecordEditor::create();
$config->removeComponentsByType(GridFieldPaginator::class);
$config->addComponent(new GridFieldConfigurablePaginator());

$fields->addFieldToTab('Root.Main',
    GridField::create("ContentBlock", 'Contentblocks', $this->ContentBlock(), $config));
```

List is empty when I use this config.

```php
$fields->addFieldToTab('Root.Main',
    GridField::create("ContentBlock", 'Contentblocks', $this->ContentBlock(), GridFieldConfig_RecordEditor::create()));
```

List is there normally.

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/438
- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/440

RoyalPulp:
I can conform this issue! Just wondered, why some GridFields were empty. 
Removing the GridFieldConfigurablePaginator made the entries reappear.

yvuyvu:
I have the same problem. We have many customer websites that we want to update to SilverStripe 5.4. Since the GridFieldConfigurablePaginator is an indispensable element on these websites.

Therefore, it is currently not an option for us to update our customer websites.

Fixing this issue would help us enormously.

michalkleiner:
[...]

PR #438, opened 2025-04-30
FIX Get GridFieldConfigurablePaginator working again.

GuySartorelli:
## Issue

- https://github.com/symbiote/silverstripe-gridfieldextensions/issues/436

### DIFF
--- a/src/GridFieldConfigurablePaginator.php
+++ b/src/GridFieldConfigurablePaginator.php
@@ -4,6 +4,7 @@ namespace Symbiote\GridFieldExtensions;
 
 use Exception;
 use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\GridField\GridField_FormAction;
 use SilverStripe\Forms\GridField\GridFieldPaginator;
@@ -45,6 +46,8 @@ class GridFieldConfigurablePaginator extends GridFieldPaginator
      */
     protected $pageSizes = array();
 
+    private bool $haveCheckedCount = false;
+
     /**
      * @param int $itemsPerPage  How many items should be displayed per page
      * @param int $pageSizes The page sizes to show in the dropdown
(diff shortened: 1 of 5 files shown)


src/GridFieldConfigurablePaginator.php
```diff
@@ -64,10 +67,21 @@ public function __construct($itemsPerPage = null, $pageSizes = null)
...
+    }
+
+    public function getTotalItems(): int
+    {
+        if (!$this->haveCheckedCount) {
+            $this->totalItems = (int) $this->getGridField()->getList()->count();
+            $this->haveCheckedCount = true;
+        }
+        return $this->totalItems;
     }
```
ChloeSartorelli, line 85: This method is introduced in the parent class 5.4 and is called from GridField.

src/GridFieldConfigurablePaginator.php
```diff
@@ -64,10 +67,21 @@ public function __construct($itemsPerPage = null, $pageSizes = null)
...
      *
      * @return int
+     * @deprecated 4.1.1 Use getTotalItems() instead.
      */
     public function getTotalRecords()
     {
-        return (int) $this->getGridField()->getList()->count();
+        Deprecation::notice('4.1.1', 'Use getTotalItems() instead.');
+        return $this->getTotalItems();
+    }
```
ChloeSartorelli, line 76: Deprecating in favour of the method in the parent class, since the purpose is the same.

src/GridFieldConfigurablePaginator.php
```diff
@@ -64,10 +67,21 @@ public function __construct($itemsPerPage = null, $pageSizes = null)
...
+
+    public function getTotalItems(): int
+    {
+        if (!$this->haveCheckedCount) {
+            $this->totalItems = (int) $this->getGridField()->getList()->count();
+            $this->haveCheckedCount = true;
+        }
```
ChloeSartorelli, line 83: Unlike in the parent class, this is relatively likely to be called at various points, so we need to check if we've counted.

Can't just rely on whether the value of `totalItems` has changed because `0` is the default and we don't want to keep counting the list if it's legitimately empty

tests/behat/features/configurable-paginator.feature
ChloeSartorelli, line 1: I tried for probably longer than I should have, but I couldn't get behat to select from the "Show 15" dropdown so that will go untested for now.

tests/behat/features/configurable-paginator.feature
```diff
@@ -0,0 +1,100 @@
+@gsat
```
emteknetnz:
```suggestion
```

emteknetnz (changes requested):
Tested and works well, just a tiny change

tests/behat/features/configurable-paginator.feature
ChloeSartorelli: Doh! Removed

emteknetnz (approved):

Issue #441, opened 2025-05-09
Error when using GridFieldConfigurablePaginator in a ModelAdmin in SilverStripe 5.4


elkebe:
### Module version(s) affected

4.1.1

### Description

Adding Configurable Paginator to model admin causes the following error in `/admin/security`:

`[Emergency] Uncaught LogicException: SilverStripe\Forms\GridField\GridFieldPageCount relies on a GridFieldPaginator to be added to the same GridField, but none are present.`

Note that this worked fine in SS5.3, broke in SS5.4

May be related to: https://github.com/symbiote/silverstripe-gridfieldextensions/issues/436

### How to reproduce

Add an extension to ModelAdmin, eg.
```
SilverStripe\Admin\ModelAdmin:
  extensions:
    - AdminPaginator
```
```
use SilverStripe\Core\Extension;
use SilverStripe\Forms\GridField\GridFieldPaginator;
use Symbiote\GridFieldExtensions\GridFieldConfigurablePaginator;

class AdminPaginator extends Extension
{
    public function updateGridFieldConfig(&$config)
    {
        $config
            ->removeComponentsByType(GridFieldPaginator::class)
            ->addComponent(
                GridFieldConfigurablePaginator::create(50, [
                    10,
                    50,
                    100,
                    200,
                    300,
                ]),
            );
    }
}
```
Go to `https://mysite/admin/security` and get error.

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

- https://github.com/symbiote/silverstripe-gridfieldextensions/pull/442

GuySartorelli:
Ohhh I see what's happened here.

In Silverstripe 5.4 because of the change that activates sudo mode for the security section, the gridfield is set to readonly mode.

`GridFieldConfigurablePaginator` is not included in the `GridField.default_readonly_components` configuration array (even though its parent class is), which means it gets excluded from the gridfield when sudo mode makes it read-only.

This YAML configuration will resolve this for you:

```yml
SilverStripe\Forms\GridField\GridField:
  default_readonly_components:
    - 'Symbiote\GridFieldExtensions\GridFieldConfigurablePaginator'
```

Debatably, one of the following should be done:

1. Add the above YAML config to this module, so that the fancy paginator is allowed for readonly gridfields.
[...]

PR #442, opened 2025-05-12
FIX Let configurable paginator work in read-only gridfields

GuySartorelli:
## Issue

- https://github.com/symbiote/silverstripe-gridfieldextensions/issues/441

### DIFF
--- a/_config/gridfieldextensions.yml
+++ b/_config/gridfieldextensions.yml
@@ -6,4 +6,8 @@ GridFieldAddNewMultiClass:
 
 SilverStripe\Forms\GridField\GridFieldDetailForm_ItemRequest:
   extensions:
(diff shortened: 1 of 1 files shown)


_config/gridfieldextensions.yml
```diff
@@ -6,4 +6,8 @@ GridFieldAddNewMultiClass:
 
 SilverStripe\Forms\GridField\GridFieldDetailForm_ItemRequest:
   extensions:
-    - Symbiote\GridFieldExtensions\Extensions\GridFieldDetailFormItemRequestExtension
+    - 'Symbiote\GridFieldExtensions\Extensions\GridFieldDetailFormItemRequestExtension'
```
ChloeSartorelli, line 9: Unrelated change - strings should ideally be always quoted in YAML.

emteknetnz (approved):
Does what it's supposed to

PR #448, opened 2025-07-23
FIX Make sure count is correct after filtering

GuySartorelli:
## Issue

- https://github.com/silverstripe/silverstripe-gridfieldextensions/issues/445

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     "require": {
         "php": "^8.1",
         "silverstripe/vendor-plugin": "^2",
-        "silverstripe/framework": "^5"
+        "silverstripe/framework": "^5.2"
     },
     "require-dev": {
         "phpunit/phpunit": "^9.6",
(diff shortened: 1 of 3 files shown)


src/GridFieldConfigurablePaginator.php
```diff
@@ -264,6 +268,10 @@ public function getManipulatedData(GridField $gridField, SS_List $dataList)
         // Assign the GridField to the class so it can be used later in the request
         $this->setGridField($gridField);
 
+        // Update item count prior to limit. This ensures filtered lists have the correct count.
+        $this->totalItems = $dataList->count();
+        $this->haveCheckedCount = true;
```
ChloeSartorelli, line 273: This is also where the `GridFieldPaginator` in framework sets the total item count.

composer.json
```diff
@@ -27,7 +27,7 @@
     "require": {
         "php": "^8.1",
         "silverstripe/vendor-plugin": "^2",
-        "silverstripe/framework": "^5"
+        "silverstripe/framework": "^5.2"
```
ChloeSartorelli, line 30: Needed so I can call `setSearchContext` in the unit test since that method is added in 5.2.0.
That's needed so I can filter the gridfield through `getManipulatedList()` and then check the count is correct post-filter.

emteknetnz (approved):
Works well

PR #450, opened 2025-07-30
ENH Optimise SQL queries to improve performance

GuySartorelli:
While not reducing the _number_ of queries, I've updated a bunch of calls to `find()` and `byID()` (which were only being used to check if a record exists) to instead use `filter()->exists()` which is more efficient.

Here we no longer need to hydrate a whole record - we just check if the record exists with that ID and if so, we know we can use that as the parent ID.

## Issue

- https://github.com/silverstripe/.github/issues/416


GuySartorelli:
CI failures are pre-existing - see https://github.com/silverstripe/silverstripe-gridfieldextensions/actions/runs/16613903155/job/47002534233

emteknetnz (approved):

PR #456, opened 2025-08-27
ENH Hide font icons from screen readers

GuySartorelli:
- Updates JS to support icons in inner span elements
- Replaces any `<i>` being used as an icon with `span` for consistency and semantic correctness
- Hides decorative icons from screen readers (usually by adding an inner `span`)
- Remove `data-icon` attribute which resulted in a mangled icon after moving the font-icon class

Without the admin PR things look very slightly different, but it's still perfectly usable so I don't think it's worth updating the constraint.

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1957

### DIFF
--- a/javascript/GridFieldExtensions.js
+++ b/javascript/GridFieldExtensions.js
@@ -523,7 +523,7 @@
 				let currState = gridField.getState();
 				let toggleState = false;
 				let pjaxTarget = $(this).attr('data-pjax-target');
-				if ($(this).hasClass('font-icon-right-dir')) {
+				if ($(this).attr('aria-expanded') === 'false') {
 					toggleState = true;
 				}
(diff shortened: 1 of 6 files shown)


javascript/GridFieldExtensions.js
```diff
@@ -523,7 +523,7 @@
 				let currState = gridField.getState();
 				let toggleState = false;
 				let pjaxTarget = $(this).attr('data-pjax-target');
-				if ($(this).hasClass('font-icon-right-dir')) {
+				if ($(this).attr('aria-expanded') === 'false') {
```
ChloeSartorelli, line 526: Ultimately the same functionality, except:
- the new condition is semantically correct ("if it's not expanded" instead of "if it has some css class")
- the icon class isn't on that element anymore so the alternative would be to find the child element first... which would be worse.

emteknetnz (approved):

PR #457, opened 2025-09-02
NEW Prepend new inline rows control

3Dgoo:
## Description
This changes new inline gridfield rows to be prepended to the start of the grid when added instead of appended to the end.

image

## Manual testing steps

Add a few rows (if no rows exist yet). Save.

Refresh the page to ensure there are existing rows loaded. Press the add row button. The new row should appear on top.

## Issues
- https://github.com/silverstripe/silverstripe-gridfieldextensions/issues/331

## Pull request checklist
- [x] CI is green

### DIFF
--- a/javascript/GridFieldExtensions.js
+++ b/javascript/GridFieldExtensions.js
@@ -195,7 +195,7 @@
 
 				tmpl.cache[this[0].id + "ss-gridfield-add-inline-template"] = tmpl(row.html());
 
-				this.find("tbody:first").append(tmpl(this[0].id + "ss-gridfield-add-inline-template", { num: num }));
+				this.find("tbody:first").prepend(tmpl(this[0].id + "ss-gridfield-add-inline-template", { num: num }));
 				this.find("tbody:first").children(".ss-gridfield-no-items").hide();
 				this.data("add-inline-num", num + 1);
 


GuySartorelli:
What is the use case for wanting configuration here? Wouldn't you want the record to just always be at the top as suggested in the linked issue?

Also, in future please do not remove the PR template. It's there for two main reasons:
1. Help you ensure you have done everything you need to do to get a pull request into a ready state (for example your commit message doesn't conform to [our guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages))
2. Help maintainers see that you have done all of those things, and give them all the context they need

3Dgoo:
I implemented as a configuration to not change the existing functionality for all users. This way it allows users to change the functionality if they want, but for all others it remains unchanged.

If you would prefer I can change the code to just make it prepend for everyone all the time. Let me know if you want me to make this change.

3Dgoo:
Would you like me to change the code to just make it prepend for every all the time?

seppzzz:
It really does not make sense (i can't think of a use case here) to append a new row.
Prepending is the way to go and should be the default case.

GuySartorelli:
> Would you like me to change the code to just make it prepend for every all the time?

Sorry I have taken a while to respond, I haven't been at work for a few weeks.

I am fine with there being configuration if you think it's valuable (i.e. if you think there's a use case where someone would _want_ to append instead of prepend) - but prepending should be the default behaviour.

javascript/GridFieldExtensions.js
```diff
@@ -195,7 +195,12 @@
 
 				tmpl.cache[this[0].id + "ss-gridfield-add-inline-template"] = tmpl(row.html());
 
-				this.find("tbody:first").append(tmpl(this[0].id + "ss-gridfield-add-inline-template", { num: num }));
+				if ($(button)?.data("prepend-rows")) {
+					this.find("tbody:first").prepend(tmpl(this[0].id + "ss-gridfield-add-inline-template", { num: num }));
+				} else {
+					this.find("tbody:first").append(tmpl(this[0].id + "ss-gridfield-add-inline-template", { num: num }));
+				}
```
ChloeSartorelli:
```suggestion
				const newItem = tmpl(this[0].id + "ss-gridfield-add-inline-template", { num: num });
				if ($(button)?.data("prepend-rows")) {
					this.find("tbody:first").prepend(newItem);
				} else {
					this.find("tbody:first").append(newItem);
				}
```
Just a small nitpick now that I'm looking at this again.

src/GridFieldAddNewInlineButton.php
```diff
@@ -32,6 +32,8 @@ class GridFieldAddNewInlineButton extends AbstractGridFieldComponent implements
 
     private $title;
 
+    private $prependRows = false;
```
ChloeSartorelli:
```suggestion
    private bool $prependRows = false;
```
New API should have explicit type hints instead of PHPDocs.
If you choose to remove the configuration, ignore this.

src/GridFieldAddNewInlineButton.php
```diff
@@ -85,6 +87,28 @@ public function setTitle($title)
         return $this;
     }
 
+    /**
+     * Gets the prepend new inline rows flag.
+     *
+     * @return bool
+     */
+    public function getPrependRows()
```
ChloeSartorelli:
```suggestion
     */
    public function getPrependRows(): bool
```
New API should have explicit type hints instead of PHPDocs.
If you choose to remove the configuration, ignore this.

src/GridFieldAddNewInlineButton.php
```diff
@@ -85,6 +87,28 @@ public function setTitle($title)
...
+        return $this->prependRows;
+    }
+
+    /**
+     * Sets the prepend new inline rows flag.
+     *
+     * @param bool $prependRows
+     * @return GridFieldAddNewInlineButton $this
+     */
+    public function setPrependRows($prependRows)
```
ChloeSartorelli:
```suggestion
     */
    public function setPrependRows(bool $prependRows): static
```
New API should have explicit type hints instead of PHPDocs.
If you choose to remove the configuration, ignore this.

GuySartorelli (changes requested):

3Dgoo:
PR updated to simply prepend instead of append. Configuration removed as requested.

GuySartorelli (approved):
LGTM
Thanks for adding this.
I'll put a quick note in the changelog.

PR #464, opened 2026-01-12
ENH Update usage of deprecated event properties

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/472

### DIFF
--- a/javascript/GridFieldExtensions.js
+++ b/javascript/GridFieldExtensions.js
@@ -172,7 +172,7 @@
 				}
 			},
 			onkeyup: function(e) {
-				if (e.keyCode == 90 && e.ctrlKey)
+				if (e.ctrlKey && e.key.toLowerCase() === "z")
 				{
 					var target = $(e.target);
 					var elementsChanged = target.data("pasteManipulatedElements");


javascript/GridFieldExtensions.js
```diff
@@ -172,7 +172,7 @@
 				}
 			},
 			onkeyup: function(e) {
-				if (e.keyCode == 90 && e.ctrlKey)
+				if (e.ctrlKey && e.key.toLowerCase() === "z")
```
emteknetnz, line 175: e.key.toLowerCase() === 'z' is preferable to e.code === 'KeyZ' as the latter may break international keyboards

ChloeSartorelli: Have validated this from several online sources - e.g. the german keyboard layout has Z and Y swapped compared to QWERTY - but pressing ctrl with the key labeled "z" still does "undo" in most applications. 

Therefore using `e.key` (which reports what the key is _labeled_) is preferred over `e.code` (which reports what position the key is in from a QWERTY perspective).

GuySartorelli (approved):
LGTM

Issue #465, opened 2026-02-08
error thrown when getContainerFieldList() returns null


xini:
### Module version(s) affected

5.1

### Description

A change introduced in https://github.com/silverstripe/silverstripe-framework/commit/aa4f7458a4c7e58344d56ea1750a23bed70ced67 results in `getContainerFieldList()` returning null.

#### 1. Add the following code to `Page.php`:

```php
use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\Forms\GridField\GridField;
use SilverStripe\Forms\GridField\GridFieldAddExistingAutocompleter;
use SilverStripe\Forms\GridField\GridFieldAddNewButton;
use SilverStripe\Forms\GridField\GridFieldConfig_RecordEditor;
use Symbiote\GridFieldExtensions\GridFieldAddNewMultiClass;
use Your\Name\Space\BaseQuestion;
use Your\Name\Space\QuestionA;
use Your\Name\Space\QuestionB;

class Page extends SiteTree
{
    ...
	
	private static $has_many = [
		'Questions' => BaseQuestion::class,
	];
	
	...

	public function getCMSFields()
	{
		$fields = parent::getCMSFields();

		// question edit config
		$questionsConfig = GridFieldConfig_RecordEditor::create();
		$questionsConfig->removeComponentsByType([
			GridFieldAddNewButton::class,
			GridFieldAddExistingAutocompleter::class,
		]);
		$itemsAdder = GridFieldAddNewMultiClass::create('before');
		$itemsAdder->setClasses([QuestionA::class, QuestionB::class]);
		$questionsConfig->addComponent($itemsAdder);
		$questionsField = GridField::create(
			'Questions',
			'Questions',
			$this->Questions(),
			$questionsConfig
		);
		$fields->addFieldsToTab(
			'Root.Questions',
			[
				$questionsField,
			]
		);

		return $fields;
	}

    ...
}
```

#### 2. Add the following classes:

```php
namespace Your\Name\Space;

use Page;
use SilverStripe\ORM\DataObject;

class BaseQuestion extends DataObject
{
    private static $table_name = 'BaseQuestion';
    private static $db = [
        'Question' => 'Varchar(255)',
        'Sort' => 'Int',
    ];
    private static $has_one = [
        'Page' => Page::class,
    ];
}
```

```php
namespace Your\Name\Space;

class QuestionA extends BaseQuestion
{
    private static $table_name = 'QuestionA';
[...]

PR #466, opened 2026-02-08
fix error when getContainerFieldList() returns null

xini:
## Description
A change introduced in https://github.com/silverstripe/silverstripe-framework/commit/aa4f7458a4c7e58344d56ea1750a23bed70ced67 results in `getContainerFieldList()` returning null sometimes. This fix catches that.

## Issues
- #465 

## Pull request checklist
- [x] CI is green

### DIFF
--- a/src/Extensions/GridFieldDetailFormItemRequestExtension.php
+++ b/src/Extensions/GridFieldDetailFormItemRequestExtension.php
@@ -28,9 +28,10 @@ class GridFieldDetailFormItemRequestExtension extends Extension
         $addMultiClassComponent = $gridFieldConfig->getComponentByType(GridFieldAddNewMultiClass::class);
         if ($addMultiClassComponent) {
             $newRecordField = static::get_new_record_field_from_actions($actions);
-            if ($newRecordField && ($container = $newRecordField->getContainerFieldList())) {
-                $container->removeByName('new-record');
-                $container->push(
+            if ($newRecordField) {
+                $containerField = $newRecordField->getContainerFieldList();
+                $containerField->removeByName('new-record');
+                $containerField->push(
                     LiteralField::create('new-record', $this->getHTMLFragment($addMultiClassComponent))
                 );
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Marking as draft pending steps to reproduce in the issue. Depending on when this actually happens, simply failing to perform any action may end up hiding a bug rather than solving one.

src/Extensions/GridFieldDetailFormItemRequestExtension.php
```diff
@@ -28,9 +28,9 @@ protected function updateFormActions(FieldList &$actions)
         $addMultiClassComponent = $gridFieldConfig->getComponentByType(GridFieldAddNewMultiClass::class);
         if ($addMultiClassComponent) {
             $newRecordField = static::get_new_record_field_from_actions($actions);
-            if ($newRecordField) {
-                $newRecordField->getContainerFieldList()->removeByName('new-record');
-                $newRecordField->getContainerFieldList()->push(
+            if ($newRecordField && ($container = $newRecordField->getContainerFieldList())) {
+                $container->removeByName('new-record');
+                $container->push(
```
ChloeSartorelli:
```suggestion
            if ($newRecordField) {
                $containerField = $newRecordField->getContainerFieldList()
                $containerField->removeByName('new-record');
                $containerField->push(
```
We're explicitly getting the field from the fieldlist, so it _does_ have a container field list to start with.

The problem was we were calling `getContainerFieldList()` _after_ removing the field. We need to capture the value returned from `getContainerFieldList()` first, instead.

GuySartorelli (changes requested):

src/Extensions/GridFieldDetailFormItemRequestExtension.php
xini: @GuySartorelli done

GuySartorelli (approved):
LGTM. Thanks for submitting this

Issue #469, opened 2026-03-19
GridFieldEditableColumns saves parent object for every grid item


xini:
### Module version(s) affected

latest 4.x and 5.x (currently 4.1.5 and 5.1.1)

### Description

This is an issue covering the previous attempts at fixing this, see https://github.com/silverstripe/silverstripe-gridfieldextensions/pull/350, https://github.com/silverstripe/silverstripe-gridfieldextensions/pull/422 and https://github.com/silverstripe/silverstripe-gridfieldextensions/pull/437.

Child objects contain a link back to the parent object (e.g. a has_one back to the parent of a has_many relation). 

When saving the parent object (say RecipePage), the child objects managed in a GridField with GridFieldEditableColumns are being saved. This in return causes the parent object (here the Recipe Page) to be saved again. 

This results in the parent object being saved as many times as there are child objects, affecting performance. Especially when other onBefore/After/During write hooks are hanging off that class (e.g. fulltext search updates, etc).

E.g.

RecipePage:
  has_many:
    Ingredients: Ingredient

Action: Save Page

current (bad) execution flow:

Write Page
Write Ingredient1
=>Write Page
Write Ingredient2
=>Write Page

good execution flow:

Write Ingredient1
Write Ingredient2
Write Page

### How to reproduce

Create a has_many from a parent page to a child object

```
class RecipePage extends Page
{
    ...
    private static $has_many = [
        'Ingredients' => Ingredient::class,
    ];
    ...
}

class RecipeIngredient extends DataObject
{
    private static $db = [
        'Title' => 'Varchar(255)',
        'Quantity' => 'Float',
        'Unit' => 'Varchar(255)'
    ];
}
```

Create a gridfield in the CMSFields of the parent page to edit the child objects using the GridFieldEditableColumns:

```
    GridField::create(
                        'Ingredients',
                        'Ingredients',
                        $this->Ingredients(),
                        GridFieldConfig_RecordEditor::create()
                            ->removeComponentsByType([
                                GridFieldAddNewButton::class,
                                GridfieldtoolbarHeader::class,
                                GridFieldFilterHeader::class,
                                GridFieldDetailForm::class,
                                GridFieldDataColumns::class,
                                GridFieldPageCount::class,
                                GridFieldPaginator::class,
                                GridfieldEditButton::class,
                                GridFieldDeleteAction::class,
                                GridField_ActionMenu::class,
                            ])
                            ->addComponents([
[...]

PR #422, opened 2024-11-25
NEW Control components written on GridFieldEditableColumns::handleSave

NightJar:
Retry #350 

Many objects contain a link back to the parent object (by nature of has_many), the same object that is being saved and triggering the GridFieldComponent to save its relational components in turn. This is often unnecessary and creates a 2N scenario for each record in the list, affecting performance. Especially when other onBefore/After/During write hooks are hanging off that class (e.g. fulltext search updates).

E.g.
```
Page:
  has_many:
    Things: Thing
```

Action: Save Page

bad execution flow:

Write Thing1
=>Write Page
Write Thing2
=>Write Page
Write Page

good execution flow:

Write Thing1
Write Thing2
Write Page

In order to have the opportunity to allow components to perform the second, this PR adds API.

PR #470, opened 2026-03-23
FIX Control which components are written on GridFieldEditableColumns:::handleSave

xini:
## Description

**This is a copy of #422, which was refused because it wasn't attached to an issue...**

This PR covers the previous attempts at fixing this, see https://github.com/silverstripe/silverstripe-gridfieldextensions/pull/350, https://github.com/silverstripe/silverstripe-gridfieldextensions/pull/422 and https://github.com/silverstripe/silverstripe-gridfieldextensions/pull/437.

Child objects contain a link back to the parent object (e.g. a has_one back to the parent of a has_many relation).

When saving the parent object (say RecipePage), the child objects managed in a GridField with GridFieldEditableColumns are being saved. This in return causes the parent object (here the Recipe Page) to be saved again.

This results in the parent object being saved as many times as there are child objects, affecting performance. Especially when other onBefore/After/During write hooks are hanging off that class (e.g. fulltext search updates, etc).

## Manual testing steps

see #469

## Issues

- #469

## Pull request checklist

- [x] CI is green

### DIFF
--- a/src/GridFieldEditableColumns.php
+++ b/src/GridFieldEditableColumns.php
@@ -52,6 +52,33 @@ class GridFieldEditableColumns extends GridFieldDataColumns implements
      */
     protected $forms = array();
 
+    /**
+     * @see DataObject::write
+     * @see DataObject::writeComponents
+     * @see DataObject::skipWriteComponents
+     *
+     * @var bool|array
+     */
+    private $skipWriteComponents = true;
+
+    /**
+     * @return bool|array
+     */
+    public function getComponentWriteConfig()
+    {
+        return $this->skipWriteComponents;
+    }
+
+    /**
+     * @param bool|array $skipConfig
+     * @return $this
+     */
+    public function setComponentWriteConfig($skipConfig)
+    {
+        $this->skipWriteComponents = $skipConfig;
+        return $this;
+    }
+
     public function getColumnContent($grid, $record, $col)
     {
         $fields = $this->getForm($grid, $record)->Fields();
@@ -169,7 +196,7 @@ class GridFieldEditableColumns extends GridFieldDataColumns implements
                 $extra = array_intersect_key($form->getData() ?? [], (array) $list->getExtraFields());
             }
 
-            $item->write(false, false, false, true);
+            $item->write(false, false, false, $this->getComponentWriteConfig());
             $list->add($item, $extra);
         }
     }


xini:
The CI errors are because of https://github.com/silverstripe/silverstripe-graphql/issues/646.

xini:
Hi @GuySartorelli,
Can you please have a look at this?
Cheers

xini:
Hello? @GuySartorelli ? @emteknetnz ?

michalkleiner:
@xini how is this used then? It seems the addition here just allows some other mechanism to skip the writes, but on its own doesn't do that. Is that correct?

xini:
@michalkleiner that's correct, it maintains backward compatibility, although IMHO most cases would need this to be set to `false`.

Imagine you have a recipe page with ingredients that you inline-edit in a gridfield. With the default settings, because there is a has_one on the ingredient to the recipe, every time an ingredient is saved, the recipe is saved as well. Now when you have 20 ingredients and save the recipe page, for each ingredient, the recipe is saved again. 

Unless the ingredients themselves have other components attached (like for example an image), it is not required/desired, that the components of the ingredients are saved when the ingredient is saved.

Usually when using inline editing, the child objects are simple, without any other objects attached. Therefore, I would argue that setting the default of this to `false` would make more sense. But for backward compatibility, this is set to `true` for now.

xini:
@michalkleiner Sorry, to answer your question, how this is used: When adding the editable columns, you set the component write config to `false`:

```
GridField::create(
	'IngredientItems',
	'Items',
	$this->IngredientItems(),
	GridFieldConfig_RecordEditor::create()
		...
		->addComponents([
			$editable = GridFieldEditableColumns::create(),
			...
		])
);
$editable->setDisplayFields([
        ...
	])->setComponentWriteConfig(
		false
	);
```

michalkleiner:
Looks ok to me, a safe change not changing the behaviour. What do you think @GuySartorelli?

michalkleiner (approved):

xini:
Thank you @michalkleiner !

xini:
HELLO?!? @GuySartorelli @emteknetnz

michalkleiner:
@xini sorry for all the delays on this. I see there were other PRs dealing with the same issue, so this is a valuable addition that will help people. Since it is not a fix, I've changed the target branch to 4 and it will be in the next minor version release.

xini:
@michalkleiner That's going to be October! I very much disagree. It is a fix without which certain sites simply don't work. Is this how Silverstripe Ltd encourages community contributions?!

NightJar:
> @michalkleiner That's going to be October! I very much disagree. It is a fix without which certain sites simply don't work. Is this how Silverstripe Ltd encourages community contributions?!

Having already tried this exact change, _twice_...

Yes.

lozcalver:
Michal is right that this is an enhancement and not a fix, so technically this should go into the next minor release. That said, there is sometimes a grey area of when something becomes a _fix_ for poor performance instead of a performance _enhancement_, and this is a very low-risk & opt-in change, so it could be argued that this is suitable for a patch release.

I’m happy to use the “spirit” of the semver rules rather than the rules as-written. Primarily because we aren’t able to just tag a minor/major release whenever we feel like it (which is what I interpret semver as being primarily designed for). But that’s just my personal position.

One note: this is targeting `4.x` (which is for CMS5) but CMS5 is no longer receiving minor releases. I don’t know if that policy extends out to modules or not, but this may need to be retargeted at either `5` or `5.2`?

xini:
Thanks, @lozcalver. I agree that the current release rules don't really align with semver (especially for non-core modules). But that's a topic for another day. 😉

> One note: this is targeting 4.x (which is for CMS5) but CMS5 is no longer receiving minor releases. I don’t know if that policy extends out to modules or not, but this may need to be retargeted at either 5 or 5.2?

This issue and PRs to fix it have been around for almost 4 years...

lozcalver:
Current status AFAICT:

- Even if this is merged into `4.1`, it won’t be auto-tagged because the 4.x CI runs [are currently failing](https://github.com/silverstripe/silverstripe-gridfieldextensions/actions/runs/25290632221/job/74141830403#step:6:395). The failures only started recently, I don’t know what’s happened there but the release tagging action won’t be triggered while they’re failing.
- If it’s merged into `4`, I don’t know if it’ll ever be tagged against that release line as SS5 is in limited support (though it would be merged up so would be released eventually, see below)
- The change _should_ still be automatically merged up to 5.x as the merge up jobs [are running okay](https://github.com/silverstripe/silverstripe-gridfieldextensions/actions/runs/25579211908/job/75093790395#step:2:797), and I don’t _think_ they depend on the CI runs passing
- The 5.x CI runs do seem to be passing, so it should be tagged against a 5.x release once it gets merged up. I assume – but haven’t checked – it would be merged into `5`, not `5.2`, so would be in the next minor release.

Sorry I can’t be of more help, I’m not involved in Silverstripe open source stuff much these days.

NightJar:
Due to BC constraints to qualify for minor/patch release, policy used to be more liberal in releasing "minor" modules when ready. CMS "product" (whole software collection e.g. "Silverstripe CMS 6.2.0) releases were stepped to quarter. I see this has this changed (product minor releases are now half yearly).

A minor release on this module would not (should not) affect the current release of the CMS as a whole, and the documentation makes no allowance for this kind of update - only whole product releases. As there is no policy directly addressing its release, I don't see what aversion there could be to make such an update wait half a year.

michalkleiner:
Acknowledging the performance issue this solves and how long it's been around, the low impact and opt-in nature of the change, I'm making the call to classify this as a fix.

I've updated the commit message to reflect that and retargeted at 5.0 for CMS 6 since CMS 5 is now in partial support and only receiving fixes for high severity security vulnerabilities.

It will be merged up automatically to 5.1, 5.2 and 5.

For CMS 5 there are existing ways how to use the fix from a custom fork.

lozcalver:
Thanks @michalkleiner!

michalkleiner:
I probably shouldn't have retargeted to 5.0, but keep it at 5.2, as I may have misunderstood the merge-up mechanism. What's your understanding @lozcalver ?

lozcalver:
Looking at the merge-up action I think you’re right, looks like it only merges up from the _most recent_ minor branch to the major branch. I think it’ll need to be manually merged up 😕

emteknetnz:
Have manually merged up to 5.1 and 5.2 -- has auto tagged -- [5.1.2](https://github.com/silverstripe/silverstripe-gridfieldextensions/releases/tag/5.1.2) -- [5.2.1](https://github.com/silverstripe/silverstripe-gridfieldextensions/releases/tag/5.2.1)

